PR #112, opened 2017-11-01
Fix example

mmorearty:
Fixed the example to call `addSourceFileFromStructure` instead of `addSourceFile`.


dsherret:
Thanks once again, Mike!

PR #190, opened 2017-12-22
Support a range of TypeScript compiler versions

### LINKED ISSUE #144: Support a range of TypeScript compiler versions
It would be nice to be able to support a range of typescript compiler versions instead of a specific one. I'm not sure how easy this would be because there are routinely breaking changes in the compiler, but perhaps all the tests could be run against multiple typescript versions on the CI.

vogloblinsky:
fix #144 

- add TypeScript multiple versions tests using npm scripts
- downgrade TypeScript version dependency


dsherret:
Excellent. Thanks very much @vogloblinsky!

Issue #204, opened 2018-01-08
Add missing Statement classes


dicarlo2:
Hi @dsherret, first off, thanks for creating a great library, it's proven to be invaluable for a project I'm currently working on. There are some missing wrappers that I'd like to see in the library. I need this rather urgently, so I'll likely publish my own package while waiting for reviews/release, but I'd love to contribute back to the library. I don't see contributor guidelines, so I'm going to just mimic some of your previous commits.

dsherret:
Sounds great! Yeah, I was thinking last night about how I can make contributing easier. I'm going to devote some time to that over the next few weeks and clean up some of the internals in the library so things are more straightforward.

dsherret:
@dicarlo2 I'll do a minor release when I get home tonight.

dicarlo2:
With these prs all statements should be there except `WithStatement`, honestly not going to bother because it's not valid typescript or strict javascript, so it basically shouldn't ever show up in a program.

dsherret:
@dicarlo2 I was feeling up for it so I went ahead and added `WithStatement`. I'm thinking perhaps someone in the future might want to write a script that automatically removes all `WithStatements` from an application.

dsherret:
I believe this is implemented now. Closing.

dicarlo2:
Thanks for the quick reviews!

PR #208, opened 2018-01-09
feat: #204 - Add IterationStatement, Do, ForIn, ForOf, For, While

dicarlo2:
Heading out for a bit, wanted to get this up in the meantime. I'll fix the merge conflicts and any review comments once I return.


src/tests/compiler/statement/doStatementTests.ts
```diff
@@ -0,0 +1,26 @@
+import * as ts from "typescript";
+import {expect} from "chai";
+import {DoStatement} from "./../../../compiler";
+import {getInfoFromText} from "./../testHelpers";
+
+function getInfoFromTextWithDoStatement(text: string) {
```
dsherret: I opened #209 because this would be useful to have in a reusable function.  I'll work on that one.

src/compiler/base/AwaitableNode.ts
```diff
@@ -0,0 +1,95 @@
...
+import {Node} from "./../common";
+import {NamedNode} from "./../base";
+
+export type AwaitableNodeExtensionType = Node<ts.Node & { awaitModifier?: ts.AwaitKeywordToken; }>;
+
+export interface AwaitableNode {
+    /**
+     * If it's a generator function.
+     */
+    isAwait(): boolean;
```
dsherret: What about `isAwaited` and `setIsAwaited`?

src/compiler/base/AwaitableNode.ts
```diff
@@ -0,0 +1,95 @@
...
+import {callBaseFill} from "./../callBaseFill";
+import {insertIntoParent, removeChildrenWithFormatting, FormattingKind} from "./../../manipulation";
+import {Node} from "./../common";
+import {NamedNode} from "./../base";
+
+export type AwaitableNodeExtensionType = Node<ts.Node & { awaitModifier?: ts.AwaitKeywordToken; }>;
+
+export interface AwaitableNode {
+    /**
+     * If it's a generator function.
```
dsherret: Needs updating.

src/compiler/base/AwaitableNode.ts
```diff
@@ -0,0 +1,95 @@
...
+     */
+    isAwait(): boolean;
+    /**
+     * Gets the await token or undefined if none exists.
+     */
+    getAwaitKeywordToken(): Node<ts.AwaitKeywordToken> | undefined;
```
dsherret: I think we should drop `Token` here—`getAwaitKeyword` and `getAwaitKeywordOrThrow`.

Other places like `ReadonlyableNode` will use `getReadonlyKeyword()` instead of including `Token` (perhaps that should be `getReadonlyToken()`, but not a big deal).

Actually, it's kind of strange that in the compiler API, the only token with the word `Keyword` in it is `AwaitKeywordToken`.

src/compiler/base/AwaitableNode.ts
```diff
@@ -0,0 +1,95 @@
...
+                insertIntoParent({
+                    insertPos: info.pos,
+                    childIndex: info.childIndex,
+                    insertItemsCount: 1,
+                    parent: this,
+                    newText: "await "
+                });
+            }
+            else {
+                removeChildrenWithFormatting({
```
dsherret: Use this:

```ts
removeChildren({
    children: [awaitModifier],
    removePrecedingSpaces: true
});
```

I need to do some refactoring around `removeChildrenWithFormatting` and this code won't cause a space to be inserted between the keyword and the parenthesis.

src/compiler/base/AwaitableNode.ts
```diff
@@ -0,0 +1,95 @@
...
+        setIsAwait(value: boolean) {
+            const awaitModifier = this.getAwaitKeywordToken();
+            const isSet = awaitModifier != null;
+
+            if (isSet === value)
+                return this;
+
+            if (awaitModifier == null) {
+                const info = getAwaitInsertInfo(this);
+                insertIntoParent({
```
dsherret: I believe this will fail when there's no space between `for` and the parenthesis. ex. `for(` -> `for(await `.

Instead use `forKeyword.getEnd()` for the position and `newText: " await"`. Then use `removePrecedingSpaces: true` below (as shown).

Generally, I will try to add spaces at the beginning instead of the end for this reason and because the spaces preceding a node are included in their position (somewhat of a factor).

dicarlo2: Good catch!

src/compiler/statement/ForStatement.ts
```diff
@@ -0,0 +1,35 @@
...
+import {IterationStatement} from "./IterationStatement";
+import {VariableDeclarationList} from "./VariableDeclarationList";
+
+export const ForStatementBase = IterationStatement;
+export class ForStatement extends ForStatementBase<ts.ForStatement> {
+
+    /**
+     * Gets this for statement's initializer.
+     */
+    getInitializer() {
```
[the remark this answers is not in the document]
dicarlo2: 👍

src/compiler/statement/VariableDeclarationList.ts
```diff
@@ -0,0 +1,146 @@
...
+import {Node} from "./../common";
+import {VariableDeclarationListStructure, VariableDeclarationStructure} from "./../../structures";
+import {ExportableNode, ModifierableNode, AmbientableNode, JSDocableNode, ChildOrderableNode} from "./../base";
+import {NamespaceChildableNode} from "./../namespace";
+import {callBaseFill} from "./../callBaseFill";
+import {VariableDeclaration} from "./VariableDeclaration";
+import {VariableDeclarationType} from "./VariableDeclarationType";
+
+export const VariableDeclarationListBase = ChildOrderableNode(ModifierableNode(Node));
+export class VariableDeclarationList extends VariableDeclarationListBase<ts.VariableDeclarationList> {
```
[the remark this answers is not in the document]
dicarlo2: Yup, wanted to make sure this change didn't cause a semver major bump.

src/tests/compiler/statement/doStatementTests.ts
```diff
@@ -0,0 +1,26 @@
...
+    const obj = getInfoFromText(text);
+    const doStatement = (
+        obj.sourceFile.getFirstDescendantByKindOrThrow(ts.SyntaxKind.DoStatement)
+    ) as DoStatement;
+    return {...obj, doStatement};
+}
+
+describe(nameof(DoStatement), () => {
+    const statement = "do {}";
```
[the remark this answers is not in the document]
dicarlo2: Wow, it seems so obvious now, but I was just puzzled as to what the expression node was supposed to be. I tried random things like `do (x > 0) {}` because for some reason I thought a `do {} while (x > 0)` would be a combination of a `DoStatement` and a `WhileStatement`.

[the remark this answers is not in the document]
dsherret: Haha, I've done something similar a few times before too :)

PR #210, opened 2018-01-09
feat #204: Add Break, Continue, Default, Labeled and Switch Statements

dicarlo2:
One thing I couldn't figure out how to resolve is the double newline when removing the last statement of a `CaseClause` or `DefaultClause` - see `statementedNodeTests.ts`.


src/compiler/statement/ClausedNode.ts
```diff
@@ -0,0 +1,78 @@
...
+import {Constructor} from "./../../Constructor";
+import * as errors from "./../../errors";
+import {verifyAndGetIndex, removeClausedNodeChildren} from "./../../manipulation";
+import {TypeGuards} from "./../../utils";
+import {Node} from "./../common";
+import {VariableDeclarationType} from "./VariableDeclarationType";
+
+export type ClausedNodeExtensionType = Node<ts.CaseBlock>;
+
+export interface ClausedNode {
```
dsherret: Does this need to be a mixin? Maybe we should just move all this directly onto `CaseBlock` then make these methods also available on `SwitchStatement`? I can't think of anything else in the language that would need this.

src/compiler/statement/StatementedNode.ts
```diff
@@ -421,6 +421,10 @@ export function StatementedNode<T extends Constructor<StatementedNodeExtensionTy
             let statements: ts.NodeArray<ts.Statement>;
             if (TypeGuards.isSourceFile(this))
```
dsherret: Does it work to do this?

```ts
if (TypeGuards.isSourceFile(this) || TypeGuards.isCaseClause(this) || TypeGuards.isDefaultClase(this))
    statements = this.compilerNode.statements;
```

src/compiler/statement/SwitchStatement.ts
```diff
@@ -0,0 +1,41 @@
...
+     */
+    getCaseBlock() {
+        return this.getNodeFromCompilerNode(this.compilerNode.caseBlock) as CaseBlock;
+    }
+
+    /**
+     * Get if the switch statement is exhaustive. False means it may or may not be exhaustive. See isNotExhaustive.
+     */
+    isExhaustive() {
+        return this.compilerNode.possiblyExhaustive === true;
```
dsherret: There must be a reason the underlying node says `possiblyExhaustive`. Maybe we should call this property `isPossiblyExhaustive()`?

Also, don't worry about also having the `isNotExhaustive` method. I believe `undefined` means `false` in this case.

src/typings/globby.d.ts
```diff
@@ -1,6 +0,0 @@
-﻿declare module "globby" {
```
dsherret: I guess this must be ok to remove?

src/compiler/statement/ClausedNode.ts
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * Gets the node's clauses.
+     */
+    getClauses(): Node[];
+    /**
+     * Adds clauses.
+     * @param text - Text of the clause or clauses to add.
+     * @returns The clauses that were added.
+     */
+    addClauses(text: string): Node[];
```
dsherret: I think we need to think through how exactly clauses should be added to a switch statement.

I'm thinking maybe something like:

```ts
const switchStatement = ...;
switchStatement.insertClause(0, {
    cases: ["5", "6", "default"],
    body: "console.log(5);" // or writer => writer.writeLine(5);
});
```

Do you need this right now? Maybe we could add this in later because I think it would require more work. If not, could you remove the insert and add method from here for now and I'll open an issue?

src/compiler/statement/ClausedNode.ts
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * Removes the clause at the specified index.
+     * @param index - Index to remove the clause at.
+     */
+    removeClause(index: number): this;
+    /**
+     * Removes the clauses at the specified index range.
+     * @param indexRange - The start and end inclusive index range to remove.
+     */
+    removeClauses(indexRange: [number, number]): this;
```
dsherret: Nice. Maybe in the future I'll add this in more places.

dsherret:
I'll take a look at the issue you were mentioning in statementedNodeTests once it's merged.

src/compiler/statement/SwitchStatement.ts
dicarlo2: I interpreted it as `undefined` meant that the compiler/type checker doesn't know. Whereas `true` means definitely exhaustive and `false` means definitely not exhaustive. Thoughts?

src/compiler/statement/ClausedNode.ts
dicarlo2: Don't need right now, will remove.

src/typings/globby.d.ts
dicarlo2: No, accident, added back.

src/compiler/statement/SwitchStatement.ts
dsherret: Oh ok, I just took a look at the code and this property is only set when it's bound by the type checker.

I'm going to remove both of these after merging.

PR #211, opened 2018-01-09
feat #204: Add ThrowStatement and TryStatement

dicarlo2:


src/compiler/statement/TryStatement.ts
```diff
@@ -0,0 +1,48 @@
...
+     */
+    getCatchClauseOrThrow() {
+        return errors.throwIfNullOrUndefined(this.getCatchClause(), "Expected to find a catch clause.");
+    }
+
+    /**
+     * Gets this try statement's finally block or undefined if none exists.
+     */
+    getFinallyBlock() {
+        return this.compilerNode.finallyBlock == null || this.compilerNode.finallyBlock.getText() === ""
```
dsherret: Why return undefined when it's empty? That might lead someone to think the block doesn't currently exist.

src/compiler/statement/CatchClause.ts
```diff
@@ -0,0 +1,33 @@
...
+     * Gets this catch clause's block.
+     */
+    getBlock() {
+        return this.getNodeFromCompilerNode(this.compilerNode.block) as Block;
+    }
+
+    /**
+     * Gets this catch clause's variable declaration or undefined if none exists.
+     */
+    getVariableDeclaration() {
```
dsherret: Interesting. There's a remove method on VariableDeclaration that will throw in this scenario. I'll fix that.

src/compiler/statement/TryStatement.ts
dsherret: Oh ok, I see now. That's weird they would do that!

Issue #213, opened 2018-01-09
Add missing Expression classes


dicarlo2:

dsherret:
If you work on this, would you mind moving the `expressions` to their own folder? Right now they're in `compiler/common`, but it would be better if they were moved to `compiler/expression`.

dsherret:
@dicarlo2 can we can close this issue now? Thanks once again for all your help! You did a great job here... I barely had to give any feedback.

dicarlo2:
Yea, I'm pretty sure this is complete, if I missed any I'll open another issue.

dicarlo2:
Thanks again for the speedy reviews!

dsherret:
Cool, this is released in v5.5.

By the way, fun fact: Node.d.ts is over 15,000 lines long since there's so many new wrapped nodes... the release took 10 minutes to run in order to generate all the overloads (ex. `getFirstChildIfKind(kind: ts.SyntaxKind.Block): compiler.Block | undefined;`). I think I'm going to make the js docs a single line. Also, in the future I'm going to need to work on improving the performance so that it's a bit quicker when working on large files.

PR #215, opened 2018-01-10
feat #213: Add missing Expression classes

dicarlo2:


src/compiler/literal/template/NoSubstitutionTemplateLiteral.ts
```diff
@@ -0,0 +1,13 @@
+// import * as ts from "typescript";
```
dsherret: Took me a little bit to figure out, but this is doing something like this:

```ts
let t: `test`;
```

I'll add this back.

dicarlo2: Oh right, sorry forgot to explain why this was commented out - I think this might not be supported by all typescript versions that ts-simple-ast supports.

dsherret: Looks like we're good. This is the 2.4 release: https://github.com/Microsoft/TypeScript/blob/8b2fe136f8cac532445bb3de89abd134592172d1/lib/typescript.d.ts#L911

PR #277, opened 2018-03-16
Add type.isStringType() and type.isNumberType()

Gerrit0:
Fairly simple change, I needed some way to test for string / number types.


dsherret:
Looks awesome. Thanks a lot!

dsherret:
@Gerrit0 this is now available in 9.2.0.

PR #314, opened 2018-04-14
Add failing test for multiple file move

mitchlloyd:
I've got a project where I'm moving multiple files that have reference each other with import statements. I'm noticing that after I move a file the import specifiers aren't updated when I later move a file that it references.

I created a failing test that I shows what I'm talking about.

Is this a current limitation of the `move` functionality or am I doing it wrong?


dsherret:
Looks like a bug to me. I’ll check it out tomorrow. Also, thanks for submitting a failing test! Really helpful!

mitchlloyd:
@dsherret This test shows the issue, but I realized that the assertions aren't quite right because the source file was moved to `"/dir/NewFile.ts"`. So hopefully that doesn't throw you off.

Although this may not be a viable solution, I noticed that removing a call to `clearDirtySourceFiles` seems to fix this test.

https://github.com/mitchlloyd/ts-simple-ast/commit/71ce6a79ec23f82d9efc305f4b1f4ed85e7cfda4

dsherret:
@mitchlloyd Yeah, I noticed and yes that's the issue. Thanks! I have a fix coming really soon. Just testing out a few more things.

dsherret:
@mitchlloyd this is fixed and released in 10.3.1. Thanks very much for reporting this!

PR #341, opened 2018-06-05
ParameteredNode.removeParameter()

cancerberoSgx:
ability to remove parameters by name


dsherret:
Hey @cancerberoSgx. Sorry, some of the methods are not discoverable, but there already exists a `parameter.remove()` method. Basically, declarations have `.remove()` method and expressions (like arguments) have some things like a `removeArgument` method on the parent.

It would be best to open an issue so we can discuss it before spending the time to implement a feature.

dsherret:
Just to clarify, the current way to do this is:

```ts
functionDeclaration.getParameterOrThrow("someName").remove();
```

cancerberoSgx:
ok thanks! will take a look, sorry for the noise. BTW - just in case you missed it, could you advice me regarding the function joinStructures I had to implement in order to "join" mixin structures while implementing getStructure of concrete classes ? https://github.com/dsherret/ts-simple-ast/issues/45#issuecomment-394555918  thanks!

PR #352, opened 2018-06-28
Fixed reading relative files from tsconfig.json.

dzinxed:
Previously, when relative file paths were passed as "files" in tsconfig,
and there were multiple root directories, all combinations of rootdir
and file path was created. This was incorrect, as the purpose of having
multiple root dirs is to have your files in one of these dirs.

To fix that, TsConfigResolver now checks if the file/subdirectory exists
before adding it to the list of files/subdirectories.

Note that with the current fix, there is no assertion that the file/directory exists at all within any of
given rootDirs. This might be fine, or I could come up with something if you think it's important.

### DIFF
--- a/src/tests/utils/tsconfig/tsConfigResolverTests.ts
+++ b/src/tests/utils/tsconfig/tsConfigResolverTests.ts
@@ -173,6 +173,21 @@ describe(nameof(TsConfigResolver), () => {
             });
         });
 
+        it("should correctly parse file paths relative to rootDirs", () => {
+            const fs = new VirtualFileSystemHost();
+            fs.writeFileSync("tsconfig.json", `{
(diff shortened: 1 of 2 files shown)


dsherret:
Thanks once again for reporting and submitting a fix for this @dzinxed!

I looked into this issue some more and I feel like this is something the TypeScript compiler API should be able to handle itself, but I went through the code of `ts.parseJsonConfigFileContent` and `ts.getFileNamesFromConfigSpecs` and sadly didn't find a way. I'm thinking I shouldn't be calling it for every directory specified in the root dirs, but I can't see an easier/more reliable way at the moment. I wouldn't be surprised if there's another bug somewhere in this code.

dsherret:
Published in 12.4.1.

PR #361, opened 2018-07-12
Browser

cancerberoSgx:
Fix https://github.com/dsherret/ts-simple-ast/issues/354
 
* wrap globby require call the same way as with fs-extra so it supports the browser
 * npm run overwrite-declaration-file :  replace copy /Y with shx cp so it support non-windows OS (needed to install shelljs). 

See example: https://gist.github.com/cancerberoSgx/2151d205c4aa13a7973cf25dd8e396d6

Feedback : 

* useVirtualFileSystem will work in the browser - no need to implement FileSystemHost from scratch
* Missing documentation - I could write a couple of paragraphs (see example bottom comments) but I'm not sure in which section, perhaps in filesystem's?
* Would be good idea to export VirtualFileSystemHost so users can override a couple of them to support, for ex, indexDB or reuse it even for other techs besides the browser. Document its methods stating that async ones are based on syncs (so users override only those) and also change modifiers of some attributes (directories) from private to protected so subclasses have visibility
* I've seen some libraries in similar situation (eslint) distribute a browser bundle or provide npm run build-browser script. is easy to do but I'm not sure is worthwhile for ts-simple-ast since I need to add browserify or webpack devDependencies
* tests ? (again IMO not worthwhile )
* future : better dependency injection - DefaultFileSystemHost is required and used in some places directly (LanguageServiceHost). Probably this could be easily solved with a factory

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -655,6 +655,12 @@
         "jsbn": "~0.1.0"
       }
     },
+    "es6-object-assign": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/es6-object-assign/-/es6-object-assign-1.1.0.tgz",
+      "integrity": "sha1-wsNYJlYkfDnqEHyx5mUrb58kUjw=",
+      "dev": true
+    },
     "escape-string-regexp": {
       "version": "1.0.5",
       "resolved": "https://registry.npmjs.org/escape-string-regexp/-/escape-string-regexp-1.0.5.tgz",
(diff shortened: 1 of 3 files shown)


dsherret:
@cancerberoSgx Awesome! Thanks so much!

dsherret:
@cancerberoSgx btw, I really appreciate the cross platform change you did in the script. I had just done that lazily as a temporary solution because I didn't know the best way to do that (you seem to have picked something good).

Also, this is released now in 12.5.3. Let me know if you have any issues.

dsherret:
Btw, I'll read your feedback section on the weekend. Please ping me in case I forget for some reason, but I've set a reminder to do so.

cancerberoSgx:
@dsherret I only worried about documenting this a somehow, but I don't want to pollute current docs with this that's kind of OT. 1)  I would do the following: add this file (or similar) somewhere in the repos (https://gist.github.com/cancerberoSgx/2151d205c4aa13a7973cf25dd8e396d6) or a simillar. 2) add a section in filesystem titled "Browser support" with small paragraph and a link to that file .3) Not sure where that section could be added. What do you think ?

PR #364, opened 2018-07-16
Browser

cancerberoSgx:
marking dependencies to be ignored by browser bundlers, references: https://github.com/Microsoft/TypeScript/blob/60986adee5a96579a258a2dd896f4bd3d2da6c98/package.json#L109


lazarljubenovic:
Oh, this means it will work in StackBlitz and similar? Nice! :tada:

cancerberoSgx:
I'm not sure - I think we need to distribute main.js file (for browsers) and declare that one in "browser" property - I don't think stackbliz will run bundlers theirself - I think they just use cdns like https://unpkg.com/ . The nice thing is if we generate that in npm pubish those cdns will expose the fie automatically.

dsherret:
Thanks! 😄

Issue #207, opened 2018-01-09
Add contributing instructions


dsherret:
This is something I haven't had a need to do for the past while, but it would be extremely beneficial.

I'd also like to spend some time getting the scripts that run on the code cleaned up and bring them up to date.

PR #367, opened 2018-07-17
isRenameable, isDeclaration and Identifier getName, hasName wil throw issue 

cancerberoSgx:
* TypeGuards.isRenameable, 
* TypeGuards.isDeclaration 
* added Identifier getName
* ISSUE detected hasName throw issue 
* added ImportSpecifier.rename() since it is implementing with wrong name renameName

### DIFF
--- a/src/compiler/common/Identifier.ts
+++ b/src/compiler/common/Identifier.ts
@@ -9,7 +9,7 @@ export class Identifier extends IdentifierBase<ts.Identifier> {
     /**
      * Gets the text for the identifier.
      */
-    getText() {
+    getText(): string {
         return this.compilerNode.text;
     }
 
@@ -37,6 +37,12 @@ export class Identifier extends IdentifierBase<ts.Identifier> {
         return this.global.languageService.getDefinitions(this);
     }
 
+    /**
+     * Gets the name of what this identifier is referencing.
+     */
+    getName(): string {
+        return this.compilerNode.text;
+    }
     /**
      * Gets the implementations of the identifier.
      *
--- a/src/compiler/file/ImportSpecifier.ts
+++ b/src/compiler/file/ImportSpecifier.ts
@@ -29,6 +29,13 @@ export class ImportSpecifier extends Node<ts.ImportSpecifier> {
      * @param name - New name.
      */
     renameName(name: string) {
+        return this.rename(name);
+    }
+    /**
+     * Renames the identifier being imported.
+     * @param name - New name.
+     */
+    rename(name: string) {
         this.getNameNode().rename(name);
         return this;
     }
--- a/src/tests/utils/typeGuardsTests.ts
+++ b/src/tests/utils/typeGuardsTests.ts
@@ -63,14 +63,14 @@ describe(nameof(TypeGuards), () => {
         });
 
         xit("should have a name if it is renameable", () => {
-            //TODO: like we did for "should be renameable if it has a name" but viceversa
+            // TODO: like we did for "should be renameable if it has a name" but viceversa
         });
     });
 
     describe(nameof(TypeGuards.isRenameableNode), () => {
         it("should be renameable when it is a declaration that has name", () => {
             const { sourceFile } = getInfoFromText(`
-                function a(){}; var b = 1; class c{}; interface d{f();h:string}; 
+                function a(){}; var b = 1; class c{}; interface d{f();h:string};
                 export default function e(){};export type g = typeof b`);
             ["a", "b", "c", "d", "e", "f", "g"].forEach(name => {
                 const node = sourceFile.getDescendants().find(n => TypeGuards.isDeclaration(n) &&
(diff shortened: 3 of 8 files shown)


src/compiler/common/Identifier.ts
```diff
@@ -9,7 +9,7 @@ export class Identifier extends IdentifierBase<ts.Identifier> {
     /**
      * Gets the text for the identifier.
      */
-    getText() {
+    getText(): string {
```
dsherret: Why is this explicit type needed?

src/compiler/common/Identifier.ts
```diff
@@ -37,6 +37,12 @@ export class Identifier extends IdentifierBase<ts.Identifier> {
         return this.global.languageService.getDefinitions(this);
     }
 
+    /**
+     * Gets the name of what this identifier is referencing.
+     */
+    getName(): string {
```
dsherret: This duplicates what `getText()` does, right? I'd rather not have multiple methods for the same thing.

src/compiler/file/ImportSpecifier.ts
```diff
@@ -29,6 +29,13 @@ export class ImportSpecifier extends Node<ts.ImportSpecifier> {
      * @param name - New name.
      */
     renameName(name: string) {
+        return this.rename(name);
+    }
+    /**
+     * Renames the identifier being imported.
+     * @param name - New name.
+     */
+    rename(name: string) {
```
dsherret: It's called `renameName` so there's no confusion about what's being renamed. Some people might expect a `rename` method to rename the alias.

Perhaps we should deprecate `renameName` and instead have this `rename` method rename the alias if it exists and rename the name if the alias doesn't exist. Would that make sense?

src/tests/utils/typeGuardsTests.ts
```diff
@@ -1,47 +1,97 @@
 import { expect } from "chai";
 import { TypeGuards } from "../../utils";
 import { getInfoFromText } from "../compiler/testHelpers";
+import { SyntaxKind } from "../../typescript";
+import "chai/register-should";
```
dsherret: I think we can remove this?

src/tests/utils/typeGuardsTests.ts
```diff
@@ -1,47 +1,97 @@
...
 
 describe(nameof(TypeGuards), () => {
     // most of the code in TypeGuards is not worth the effort to test... it's auto generated from code so it should be close to correct
 
     describe(nameof(TypeGuards.hasExpression), () => {
         it("should have an expression when it's a function call", () => {
-            const {firstChild} = getInfoFromText("funcCall()");
+            const { firstChild } = getInfoFromText("funcCall()");
```
dsherret: Thanks!

src/tests/utils/typeGuardsTests.ts
```diff
@@ -1,47 +1,97 @@
...
 
         it("should not have a name when it doesn't", () => {
-            const {firstChild} = getInfoFromText("func()");
+            const { firstChild } = getInfoFromText("func()");
             expect(TypeGuards.hasName(firstChild)).to.be.false;
         });
+
+        xit("should never throw", () => { // is throwing!
+            const { sourceFile } = getInfoFromText(`export default function(){}`);
+            sourceFile.getDescendants().forEach(node => {
```
dsherret: I'll fix this in #369.

src/utils/TypeGuards.ts
```diff
@@ -1806,6 +1844,17 @@ export class TypeGuards {
...
 
+    /**
+     * Gets if the node can be renamed, this is, implements `rename()` method .
+     * @param node - Node to check.
+     */
+    static isRenameableNode(node: compiler.Node): node is compiler.Node & { rename: (name: string)=> this; } {
+        // this method is manually maintained
+        if ((node as any).rename == null)
+            return false;
+        return typeof (node as any).rename === "function";
```
dsherret: This could just be this one line.

src/utils/TypeGuards.ts
```diff
@@ -412,6 +412,44 @@ export class TypeGuards {
         return node.getKind() === SyntaxKind.DebuggerStatement;
     }
 
+    /**
+     * Gets if the node is a declaration.
+     * @param node - Node to check.
+     */
+    static isDeclaration(node: compiler.Node): node is compiler.ArgumentedNode & compiler.Node {
```
dsherret: See the warning at the top of this file. The `is` methods of this file are code generated so this will be wiped out next time the code generation runs.

dsherret:
Hey @cancerberoSgx, thanks for this, but unfortunately I'm going to have to decline these changes.

* The `isDeclaration` and `isRenameable` method should be auto-maintained by the code-generate command. That means changing up the code a bit. I'd like to work on that because it's a bit more foundational. I've opened #370 and will get to it soon.
* There's no name on an `Identifier` so I'd rather it didn't have a name. Is there a reason you wanted this?

Anyway, sorry about that. For public api changes, I recommend opening an issue and talking about it first (sorry if I take a bit to respond sometimes). In the meantime I'll get to #207.

PR #372, opened 2018-07-20
docs: Fix missing a code block end quote

zhangciwu:
Fix a markdown code block quoting problem.
Found at reading the document


dsherret:
Thanks! 😄 

I don't pay enough attention to the markdown preview to catch these mistakes.

PR #401, opened 2018-08-22
Use readonly arrays for file globs

rhys-vdw:
Just a simple type fix. `ReadonlyArray<string>` is a narrowing of `Array<string>` so should be used in any place where an argument is not mutated by library code.

I checked multimatch and globby to make sure they're not mutating their inputs.

### DIFF
--- a/src/Project.ts
+++ b/src/Project.ts
@@ -165,7 +165,7 @@ export class Project {
      * @param fileGlobs - File glob or globs to add files based on.
      * @returns The matched source files.
      */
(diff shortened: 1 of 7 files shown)


dsherret:
@rhys-vdw thanks a lot! I'm going to update other places that should do this as well.

rhys-vdw:
No problem. I've got a couple more helpers to upstream and bugs to report from my project. I'll get around to them eventually.

PR #482, opened 2018-10-28
importDeclaration.getStructure moduleSpecifier without quotes

cancerberoSgx:
fix incompatibility between ImportDeclaration.getStructure and moduleNode.addImportDeclaration - basically moduleSpecifier was quoted in getStructure()

### DIFF
--- a/src/tests/compiler/module/importDeclarationTests.ts
+++ b/src/tests/compiler/module/importDeclarationTests.ts
@@ -1,5 +1,5 @@
 ﻿import { expect } from "chai";
-import { ImportDeclaration, ModuledNode } from "../../../compiler";
+import { ImportDeclaration } from "../../../compiler";
 import * as errors from "../../../errors";
 import { Project } from "../../../Project";
 import { WriterFunction } from "../../../types";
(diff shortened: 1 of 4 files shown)


src/tests/compiler/module/importDeclarationTests.ts
```diff
@@ -707,10 +707,19 @@ describe(nameof(ImportDeclaration), () => {
...
                 defaultImport: "bar",
-                moduleSpecifier: "\'foo\'",
+                moduleSpecifier: "foo",
                 namedImports: [{ name: "test", alias: undefined }],
                 namespaceImport: undefined
             });
         });
     });
+
+    describe(nameof<ImportDeclaration>(n => n.getStructure) + " compatibiltiy with " +
```
cancerberoSgx: this test show the problem

dsherret: Thanks! I removed this though because it's superfluous. The other tests above ensure the problem won't happen and I'd prefer all tests just test one thing rather than testing multiple aspects at once.

dsherret:
Thanks a lot for the fix @cancerberoSgx! This was also incorrectly being done in export declarations so I've included that in this PR.

dsherret:
Released now in 18.0.0.

PR #483, opened 2018-10-28
GetDerivedClasses() fails with "Interfaces Extending Classes":

brianzinn:
https://www.typescriptlang.org/docs/handbook/interfaces.html

Hi, would like to add an extra guard clause to allow Interfaces to Extend classes, but not throw.


dsherret:
Hey @brianzinn, thanks for this fix! Could you also include a test that shows this code does indeed fix the problem? If not I can get around to it eventually. Thanks!

brianzinn:
I'll write a test. Cheers.

dsherret:
Really appreciate it. Thanks @brianzinn! I'll do a release for this today with some of my other changes.

brianzinn:
Great - thanks!! Thanks also for doing the release - I was wondering when it would reach NPM.  Thanks for calling me out on the test - I think default behaviour for me is laziness :)

dsherret:
Ok this is released now in 18.0.0. Let me know if you run into any issues.

Also, that's no problem! I just wanted to ensure this bug wouldn't come back in the future. Thanks again for your help :)

PR #487, opened 2018-10-30
dependencies updated - fixes npm install 

cancerberoSgx:
This is more a question than an actual PR. npm install doesn't work on a fresh git clone. this is because dependencies typescript-3.0.3: npm:typescript . If I remove this then it works. Is that the expected behavior ? 

Besides this this PR updateds all dependencies to latest version, but @types/node because a previous discussion we had regarding this one.


dsherret:
Use `yarn install`. Npm doesn't support aliases and I need those in order to test against all the different versions of the typescript compiler that are supported.

dsherret:
Also, don't worry about updating the dependencies. I'd rather leave them at the lower version.

PR #491, opened 2018-11-06
Project.applyFileTextChanges

cancerberoSgx:
For supporting https://github.com/dsherret/ts-simple-ast/pull/491

Created this branch from branch used in https://github.com/dsherret/ts-simple-ast/pull/488 - because method isFileNew() was added there so we avoid conflicts (will be merging the later with this one when change). 

The commit is just the last one 

Just wanted to start discussing the signature - open to suggestions


dsherret:
Thanks @cancerberoSgx! This is released now.

Issue #530, opened 2019-02-12
feature request: JSDocTypeExpression


Pineapples:
I'd like to request a wrapper for JSDocTypeExpression to read out the types of parameters in JSDoc tags.

Pineapples:
I've added a pull request. Seems like your dev is behind on master. Cool project to work on. Let me know what you think.

dsherret:
Thanks for the contributions @Pineapples! I just took a brief look and it looks good, but I have to go to work now. I'll merge this tonight after work.

By the way, I've updated the dev branch. Sorry about that, I just forgot to push it. Thanks for following the contributing instructions so well though :)

dsherret:
This is released in v1.1.0. Thanks again, @Pineapples!

PR #531, opened 2019-02-12
Add JSDocTypeExpression

Pineapples:
Implementation for #530


src/compiler/ast/doc/JSDocTypeExpression.ts
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * JS doc type expression node
+ */
+export class JSDocTypeExpression extends Node<ts.JSDocTypeExpression> {
+    /**
+     * Gets the type node of the JS Doc type expression
+     */
+    getTypeNode() {
+        return this.compilerNode.type;
```
dsherret: This will return the typescript compiler api node instead of the wrapped one. Use either `this._getNodeFromCompilerNode` or `this._getNodeFromCompilerNodeIfExists` (depending on whether `type` is nullable) similar to what you did in `JSDocPropertyLikeTag`.

Hmmm... I should actually combine `_getNodeFromCompilerNode` and `_getNodeFromCompilerNodeIfExists` sometime, but that's for another day.

src/tests/compiler/ast/doc/base/jsDocPropertyLikeTagTests.ts
```diff
@@ -9,18 +9,29 @@ describe(nameof(JSDocPropertyLikeTag), () => {
         return { descendant: info.sourceFile.getFirstDescendantOrThrow(TypeGuards.isJSDocPropertyLikeTag), ...info };
     }
 
+    describe(nameof<JSDocPropertyLikeTag>(d => d.getTypeExpressionNode), () => {
+        function doTest(text: string, expected: string) {
+            const { descendant } = getInfo(text);
+            expect(descendant.getTypeExpressionNode()!.getTypeNode().getText()).to.equal(expected);
```
dsherret: Could you add a test for when this doesn't exist? I usually use this same function, change `expected: string | undefined`, then change the body appropriately, but anything works.

src/tests/compiler/ast/doc/jsDocTypeExpressionTests.ts
```diff
@@ -0,0 +1,22 @@
...
+import {TypeGuards} from "../../../../utils";
+import {expect} from "chai";
+
+describe(nameof(JSDocTypeExpression), () => {
+    function getInfo(text: string) {
+        const info = getInfoFromText(text);
+        return { descendant: info.sourceFile.getFirstDescendantOrThrow(TypeGuards.isJSDocTypeExpression), ...info };
+    }
+
+    describe.only(nameof<JSDocTypeExpression>(d => d.getTypeNode), () => {
```
dsherret: Could you remove `.only` here?

src/tests/compiler/ast/doc/jsDocTypeExpressionTests.ts
```diff
@@ -0,0 +1,22 @@
...
+        function doTest(text: string, expected: string) {
+            const { descendant } = getInfo(text);
+            expect(descendant.getTypeNode().getText()).to.equal(expected);
+        }
+
+        it("should get when type is given", () => {
+            doTest("/** @param {boolean} t - String */\nfunction test() {}", "boolean");
+        });
+    });
+});
```
dsherret: Looks good! Thanks!

src/tests/compiler/ast/doc/jsDocTypeExpressionTests.ts
Pineapples: Oops..

Pineapples:
Implemented the requested changes and merged updated dev (had some conflicts). Ran the code-generate before, but didn't check in the unchanged files, was missing the other two commands.

Thanks for the feedback!

lib/code-block-writer.d.ts
```diff
@@ -1,210 +1,210 @@
-interface CodeBlockWriterOptions {
```
dsherret: I wonder why git said this file changed. I don't see any changes to line endings when I open it in a diff. Oh well, not a big deal.

src/compiler/ast/doc/base/JSDocPropertyLikeTag.ts
```diff
@@ -18,6 +21,10 @@ export interface JSDocPropertyLikeTag {
 export function JSDocPropertyLikeTag<T extends Constructor<JSDocPropertyLikeTagExtensionType>>(Base: T): Constructor<JSDocPropertyLikeTag> & T {
     return class extends Base implements JSDocPropertyLikeTag {
         // todo: more methods
+        getTypeExpression() {
```
dsherret: I shortened this method name. I've only been adding the "Node" suffix if it could be confused with something else (ex. `getType()` could be confused with getting a `Type` object and `getName()` could be confused with getting the string name, so I add the `Node` suffix in those cases)

dsherret:
Thanks so much once again! I'll do a release tonight.

Issue #532, opened 2019-02-12
feature request: Expose isBracketed on JSDocPropertyLikeTag


Pineapples:
Requesting the isBracketed boolean on a JSDocPropertyLikeTag to be exposed.

Pineapples:
Added a PR for this, let me know what you think.

PR #533, opened 2019-02-12
expose isBracketed on JSDocPropertyLikeTag

Pineapples:
Implementation for #532


dsherret:
@Pineapples thanks a lot!

Issue #535, opened 2019-02-14
feature request: Expose typeExpression for JSDocTypeTag and JSDocReturnTag


Pineapples:
Sorry to bother you again with another feature request, but I would like to have the typeExpression property for JSDocTypeTag and JSDocReturnTag exposed.

I'll submit a pull request soon.

dsherret:
For sure! I'll take a look at it after work. The jsdoc area hasn't been fully wrapped so I know it's missing stuff... sorry about that, but thanks for adding those methods. I appreciate it!

Maybe I will spend a bit of time this weekend wrapping more stuff in there.

Btw, take note that as a temporary workaround you can always fall back to the typescript compiler api node from any node by inspecting the `.compilerNode` property.

PR #536, opened 2019-02-14
Expose  typeExpression on JSDocReturnTag and JSDocTypeTag

Pineapples:
Implementation for #535


src/tests/compiler/ast/doc/jsDocTypeTagTests.ts
```diff
@@ -0,0 +1,23 @@
...
+describe.only(nameof(JSDocTypeTag), () => {
+    function getInfo(text: string) {
+        const info = getInfoFromText(text);
+        return { descendant: info.sourceFile.getFirstDescendantOrThrow(TypeGuards.isJSDocTypeTag), ...info };
+    }
+
+    describe(nameof<JSDocTypeTag>(d => d.getTypeExpression), () => {
+        it("returns undefined when typeExpression is undefined", () => {
+            const { descendant } = getInfo("/** @type */\nvar bar = 1;");
+            descendant.compilerNode.typeExpression = undefined;
```
Pineapples: Not happy about this. The Typescript type definition says it can be undefined. But I can't seem to find any way to get it actually undefined. Doesn't matter what I put behind the "@type" tag, it will always give back a value, even when it's empty.

I've put it as undefined here myself to check show that the function doesn't do anything weird when its undefined, but I don't think it can ever be undefined. Seems like this is a mistake in Typescript.

src/tests/compiler/ast/doc/jsDocTypeTagTests.ts
```diff
@@ -0,0 +1,23 @@
+import {expect} from "chai";
+import {TypeGuards, JSDocTypeTag} from '../../../../main';
+import {getInfoFromText} from '../../testHelpers';
+
+describe.only(nameof(JSDocTypeTag), () => {
```
Pineapples: I should learn to check for this..

src/tests/compiler/ast/doc/jsDocTypeTagTests.ts
dsherret: Yeah, most likely a mistake in the compiler.... that occasionally happens. Thanks for calling attention to this... I've made it return undefined if the node length is zero.

PR #538, opened 2019-02-14
added isAny() to Type

schiller-manuel:


dsherret:
Thanks @schiller-manuel! Looks great!

Seems like `isUnknown()` is missing too.

dsherret:
Also, thanks for updating the docs when making this change.

PR #540, opened 2019-02-15
Rename signature declaration

schiller-manuel:
I assume that _Signature**d**Declaration_ is a typo and should actually be _SignatureDeclaration_.
If my assumption is wrong, please close this PR.


dsherret:
Hmmm... no that's not a typo. I've been using `-ed` or `-able` endings for the mixins (there's a few exceptions like `FunctionLikeDeclaration`).

That said, there is a `ts.SignatureDeclaration` type alias in the compiler api. I'm not sure what we should do in this case, but I'm thinking we could perhaps create an additional alias and type guard for `SignatureDeclaration`? https://github.com/dsherret/ts-morph/blob/master/src/compiler/ast/aliases.ts

schiller-manuel:
ok!

Issue #542, opened 2019-02-15
wrap signature.getDeclaration


schiller-manuel:
I did not find `signature.getDeclaration` to be provided by ts-morph.
Is this functionality already somehow covered? If not, I would like to contribute this feature.

dsherret:
Hmm yeah looks like `JSDocSignature` would need to be implemented at the same time. Thanks also for this!

Issue #541, opened 2019-02-15
wrap typeChecker.getResolvedSignature


schiller-manuel:
I did not find `typeChecker.getResolvedSignature` to be wrapped by ts-morph.
Is this functionality already somehow covered? If not, I would like to contribute this feature.

dsherret:
Sure, feel free. Thanks!

PR #544, opened 2019-02-16
features for #541 and #542

schiller-manuel:
If you'd like to see other/more tests, let me know.
Is one PR for this OK, or do you want separate ones?


src/factories/CompilerFactory.ts
```diff
@@ -505,6 +506,16 @@ export class CompilerFactory {
         return this.jsDocTagInfoCache.getOrCreate(jsDocTagInfo, () => new JSDocTagInfo(jsDocTagInfo));
     }
 
+    /**
+     * Gets a wrapped signature declaration from a compiler signature declaration.
+     * @param compilerObject - Compiler signature declaration.
+     */
+    getSignatureDeclaration(compilerObject: ts.SignatureDeclaration): SignaturedDeclaration {
```
schiller-manuel: `SignaturedDeclaration` is probably wrong.
Which type is the right one?

dsherret:
Thanks a lot for this! We'll need to implement `#547` before merging this though. I'll work on that now.

src/factories/CompilerFactory.ts
dsherret: There's no need for this specific method as this functionality already exists in the more general method `CompilerFactory#getNodeFromCompilerNode`. (See my comment in `Signature.ts`)

src/compiler/symbols/Signature.ts
```diff
@@ -66,4 +66,12 @@ export class Signature {
         const tags = this.compilerSignature.getJsDocTags();
         return tags.map(t => this._context.compilerFactory.getJSDocTagInfo(t));
     }
+
+    /**
+     * Gets the signature's declaration.
+     */
+    getDeclaration(): SignaturedDeclaration {
```
dsherret: The explicit type can be removed here in order to leave this up to the typescript compiler based on the type of `compilerSignatureDeclaration`. For example:

```ts
    getDeclaration() {
        const { compilerFactory } = this._context;
        const compilerSignatureDeclaration = this.compilerSignature.getDeclaration();
        return compilerFactory.getNodeFromCompilerNode(compilerSignatureDeclaration, compilerFactory.getSourceFileForNode(compilerSignatureDeclaration));
    }
```

src/compiler/tools/TypeChecker.ts
```diff
@@ -179,6 +180,27 @@ export class TypeChecker {
...
     }
 
+    /**
+     * Gets the resolved signature from a node.
+     * @param node - Node to get the signature from.
+     * @returns undefined if signature cannot be resolved
+     */
+    getResolvedSignature(node: Node<ts.CallLikeExpression>): Signature | undefined {
+        const resolvedSignature = this.compilerObject.getResolvedSignature(node.compilerNode);
+        if (!resolvedSignature || !resolvedSignature.declaration) {
```
dsherret: Why return undefined when `.declaration` doesn't exist? Is that when it's the "unknown signature"? We should probably copy what's done in the compiler API here and not return undefined.

src/compiler/tools/TypeChecker.ts
```diff
@@ -179,6 +180,27 @@ export class TypeChecker {
         return symbol == null ? undefined : this._context.compilerFactory.getSymbol(symbol);
     }
 
+    /**
+     * Gets the resolved signature from a node.
+     * @param node - Node to get the signature from.
+     * @returns undefined if signature cannot be resolved
+     */
+    getResolvedSignature(node: Node<ts.CallLikeExpression>): Signature | undefined {
```
dsherret: I just added the `CallLikeExpression` alias in the dev branch. You can use it here instead of `Node<ts.CallLikeExpression>`.

src/compiler/tools/TypeChecker.ts
```diff
@@ -179,6 +180,27 @@ export class TypeChecker {
...
+        }
+        return this._context.compilerFactory.getSignature(resolvedSignature);
+    }
+
+    /**
+     * Gets the resolved signature from a node. Throws an error if signature cannot be resolved.
+     * @param node - Node to get the signature from.
+     */
+    getResolvedSignatureOrThrow(node: Node<ts.CallLikeExpression>): Signature {
+        return errors.throwIfNullOrUndefined(this.getResolvedSignature(node), "signature could not be resolved");
```
dsherret:
```suggestion
        return errors.throwIfNullOrUndefined(this.getResolvedSignature(node), "Signature could not be resolved");
```

src/compiler/tools/TypeChecker.ts
```diff
@@ -179,6 +180,27 @@ export class TypeChecker {
         return symbol == null ? undefined : this._context.compilerFactory.getSymbol(symbol);
     }
 
+    /**
+     * Gets the resolved signature from a node.
+     * @param node - Node to get the signature from.
+     * @returns undefined if signature cannot be resolved
+     */
+    getResolvedSignature(node: Node<ts.CallLikeExpression>): Signature | undefined {
+        const resolvedSignature = this.compilerObject.getResolvedSignature(node.compilerNode);
```
dsherret: I wonder what the other arguments are used for. It doesn't seem like a big deal to leave them out for now.

schiller-manuel:
I splitted this PR into #549 and #550

src/compiler/tools/TypeChecker.ts
schiller-manuel: I checked typescripts internal usage of this method but didn't make much sense of it, so I left them out

src/compiler/tools/TypeChecker.ts
schiller-manuel: fixed this in #550

src/compiler/tools/TypeChecker.ts
schiller-manuel: fixed this in #550

src/compiler/tools/TypeChecker.ts
schiller-manuel: let's continue discussion in #550

src/compiler/symbols/Signature.ts
schiller-manuel: fixed this in #549

PR #549, opened 2019-02-19
feat #542: Add Signature.getDeclaration()

schiller-manuel:


dsherret:
Thanks again! I added `JSDocFunctionType` so that all the types in the return type implement `SignaturedDeclaration`.

PR #550, opened 2019-02-19
feat #541: Add TypeChecker.getResolvedSignature()

schiller-manuel:


dsherret:
Hmmm... seems like the compiler api is lying about `.getDeclaration()` never returning undefined.

src/compiler/tools/TypeChecker.ts
```diff
@@ -179,6 +181,27 @@ export class TypeChecker {
...
     }
 
+    /**
+     * Gets the resolved signature from a node.
+     * @param node - Node to get the signature from.
+     * @returns undefined if signature cannot be resolved
+     */
+    getResolvedSignature(node: CallLikeExpression): Signature | undefined {
+        const resolvedSignature = this.compilerObject.getResolvedSignature(node.compilerNode);
+        if (!resolvedSignature || !resolvedSignature.declaration) {
```
dsherret: After looking into this more, I think what you did here is more sane behaviour.

dsherret:
Thanks again! 😄

I'll try to do a release later tonight.

Issue #586, opened 2019-03-27
Add Node.getEndLinePos()


cancerberoSgx:
Node has `getStartLinePos(includeJsDocComments?: boolean)`, `getStartLineNumber(includeJsDocComments?: boolean)` and `getEndLineNumber()` methods, but no `getEndLinePos()` . Can I add it ? Or is there an obvious  way of getting it that I'm not seeing? (currently using `ts.getLineAndCharacterOfPosition()`)

Thanks

dsherret:
No, doesn't exist at the moment. What's the use case for this?

cancerberoSgx:
I'm selecting the node's text in monaco-editor. I need lineNum and character both on start and end, that's the only one missing :(

dsherret:
I recommend using `sourceFile.getLineNumberAtPos(node.getEnd())` and `sourceFile.getLengthFromLineStartAtPos(node.getEnd())` for now.

We should consider possibly a `sourceFile.getLineAndColumnAtPos(pos: number): { line: number; column: number; }` maybe and deprecate `SourceFile#getLineNumberAtPos(pos: number)`.

The problem I see with `getEndLinePos` is that it's ambiguous as to whether a method with this name would return the start position of the line or the end. `getStartLinePos` is useful for quickly telling if two nodes are on the same line (doesn't require getting the line number, which is slower).

dsherret:
@cancerberoSgx if you'd like you can submit a PR for the `getLineAndColumnAtPos(pos: number): { line: number; column: number; }` method (and deprecate `getLineNumberAtPos`). Would that cover your use case?

cancerberoSgx:
yes, will do, Do you want me to also  take care of the "deprecation" part? or should I leave that part to you? Thanks

dsherret:
If you'd like. Up to you! This can go into 2.0 so it's ok to be a breaking change.

dsherret:
Thanks! 😄

cancerberoSgx:
OK, I just want to make sure I'm correct assuming It's OK that I just remove the method, right ? I'm asking because some people first tag the method with @deprecated but leave it a couple of releases before actually breaking. I'm 99% sure this is not the case, but would like to verify, thanks!

dsherret:
I usually just add `@deprecated` when I don't want to cause a major release, but I'm doing a major release anyway so we might as well remove it at this time.

dsherret:
Closing in favour of #587.

PR #587, opened 2019-03-27
SourceFile#getLineAndColumnAtPos

cancerberoSgx:
* https://github.com/dsherret/ts-morph/issues/586
 * adds SourceFile#getLineAndColumnAtPos 
 * adds tests for SourceFile#getLineAndColumnAtPo
 * removes SourceFile#getLineNumberAtPos
 * adds docs in breaking-changes.md

### DIFF
--- a/breaking-changes.md
+++ b/breaking-changes.md
@@ -4,14 +4,6 @@ View [CHANGELOG.md](CHANGELOG.md) for more detail on releases. This file is only
 
 ## Version 2
 
-### Removed SourceFile#getLineNumberAtPos()
-
-Removed `SourceFile`'s method `getLineNumberAtPos` in favor of method `getLineAndColumnAtPos` which returns an object with both line number and column number at given position: 
-
-```ts
-const { line, column } = sourceFile.getLineAndColumnAtPos(position);
-```
-
 ### Removed default export
 
 For a while now it's been recommended to use the named export instead of the default export for the project. From now on, only the named export can be used:
@@ -79,6 +71,14 @@ These two structures are now differentiated based on their new `.kind` property.
 ### `ModuledNode#getExportedDeclarations()` now returns a map
 
 So instead of just returning an array of nodes, it now returns a map. The key is the name it was exported on and the value is an array of declarations for that value. This will make it much easier to identify the name a node was exported on.
+
+### Removed SourceFile#getLineNumberAtPos()
+
+Removed `SourceFile`'s method `getLineNumberAtPos` in favor of method `getLineAndColumnAtPos` which returns an object with both line number and column number at given position: 
+
+```ts
+const { line, column } = sourceFile.getLineAndColumnAtPos(position);
+```
 
 ## Version 1
 
--- a/breaking-changes.md
+++ b/breaking-changes.md
@@ -4,6 +4,14 @@ View [CHANGELOG.md](CHANGELOG.md) for more detail on releases. This file is only
 
 ## Version 2
 
+### Removed SourceFile#getLineNumberAtPos()
+
+Removed `SourceFile`'s method `getLineNumberAtPos` in favor of method `getLineAndColumnAtPos` which returns an object with both line number and column number at given position: 
+
+```ts
+const { line, column } = sourceFile.getLineAndColumnAtPos(position);
+```
+
 ### Removed default export
 
 For a while now it's been recommended to use the named export instead of the default export for the project. From now on, only the named export can be used:
--- a/src/compiler/ast/module/SourceFile.ts
+++ b/src/compiler/ast/module/SourceFile.ts
@@ -142,14 +142,6 @@ export class SourceFile extends SourceFileBase<ts.SourceFile> {
     }
 
     /**
-     * Gets the line number at the provided position.
-     * @param pos - Position
-     */
-    getLineNumberAtPos(pos: number) {
-        return StringUtils.getLineNumberAtPos(this.getFullText(), pos);
-    }
-
-    /**
      * Gets the line number and column number at the provided position.
      * @param pos - Position
      */
--- a/src/compiler/ast/module/SourceFile.ts
+++ b/src/compiler/ast/module/SourceFile.ts
@@ -142,14 +142,14 @@ export class SourceFile extends SourceFileBase<ts.SourceFile> {
     }
 
     /**
-     * Gets the line number and column number at the provided position.
-     * @param pos - Position
+     * Gets the line number and column number at the provided position, both as 1-indexed numbers.
+     * @param pos - Position in this source file from which to obtain the line number and column number.
      */
     getLineAndColumnAtPos(pos: number) {
         const fullText = this.getFullText();
         return {
             line: StringUtils.getLineNumberAtPos(fullText, pos),
-            column: StringUtils.getLengthFromLineStartAtPos(fullText, pos)
+            column: StringUtils.getLengthFromLineStartAtPos(fullText, pos) + 1
         };
     }
 
--- a/src/compiler/ast/module/SourceFile.ts
+++ b/src/compiler/ast/module/SourceFile.ts
@@ -150,6 +150,18 @@ export class SourceFile extends SourceFileBase<ts.SourceFile> {
     }
 
     /**
+     * Gets the line number and column number at the provided position.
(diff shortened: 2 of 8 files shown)


src/tests/compiler/ast/module/sourceFileTests.ts
```diff
@@ -1405,4 +1405,79 @@ const t = 5;`;
...
+            test(code, code.indexOf("C implements") - 1, { line: 16, column: 3});
+            test(code, code.indexOf("/* line ending with spaces */ m1(){      ") + "/* line ending with spaces */ m1(){      ".length,
+                { line: 18, column: "/* line ending with spaces */ m1(){      ".length});
+            test(code, code.indexOf("// tail") + "// tail".length, { line: 22, column: "// tail".length});
+            test(code, code.indexOf("// tail") + "// tail".length + 1, { line: 23, column: 0});
+        });
+
+        it.skip("should throw when position is length + 1", () => {
+            test(`var a = 1`, `var a = 1`.length + 1, "throw");
+            test(``, 1, "throw");
```
cancerberoSgx: @dsherret this is the main doubt I have, is createSourceFile adding extra characters since this should throw ? If so, do you know if I can disable that behavior so it throws?

cancerberoSgx:
@dsherret Should we add getLineAndColumn in Node for start and end ? if so, how should be named ?  getLineAndColumnAtStart or getStartLineAndColumn ?

src/tests/compiler/ast/module/sourceFileTests.ts
dsherret: I'm not sure. Did you try debugging to check? I can look at it later if not.

src/tests/compiler/ast/module/sourceFileTests.ts
```diff
@@ -1405,4 +1405,79 @@ const t = 5;`;
...
+        });
+        it("should return line and column numbers at given position in empty first line source files", () => {
+            const code = `\ninterface I { m1(): void }`;
+            test(code, 0, { line: 1, column: 0 });
+            test(code, 1, { line: 2, column: 0 });
+            test(code, code.length, { line: 2, column: code.length - 1 });
+        });
+        it("should return line and column numbers at given position in multiple lines and comments source file", () => {
+            const code = `
+/**
```
dsherret: Could you cut down this test to 2-3 lines? It's needlessly long which increases the burden when changing the code. Don't worry about testing for spaces or no spaces.

tslint.json
```diff
@@ -5,6 +5,7 @@
             true,
             "statements"
         ],
+        "no-trailing-whitespace": [true, "ignore-template-strings"],
```
dsherret: Could you revert this? I'd like to know about trailing whitespaces in template strings.

src/tests/compiler/ast/module/sourceFileTests.ts
```diff
@@ -1123,7 +1123,7 @@ function myFunction(param: MyClass) {
 
             const referencing = sourceFile.getReferencingNodesInOtherSourceFiles();
             expect(referencing.map(r => r.getText()).sort()).to.deep.equal([...[...file1.getImportDeclarations(),
-                ...file2.getImportDeclarations(), ...file3.getExportDeclarations()].map(d => d.getText()),
```
dsherret: Is your editor doing this auto-formatting? Could you revert it?

src/compiler/tools/results/Diagnostic.ts
```diff
@@ -62,7 +62,7 @@ export class Diagnostic<TCompilerObject extends ts.Diagnostic = ts.Diagnostic> {
         const start = this.getStart();
         if (sourceFile == null || start == null)
             return undefined;
-        return sourceFile.getLineNumberAtPos(start);
+        return StringUtils.getLineNumberAtPos(sourceFile.getFullText(), start);
```
dsherret: +1

src/compiler/ast/module/SourceFile.ts
```diff
@@ -142,11 +142,15 @@ export class SourceFile extends SourceFileBase<ts.SourceFile> {
...
 
     /**
-     * Gets the line number at the provided position.
+     * Gets the line number and column number at the provided position.
      * @param pos - Position
      */
-    getLineNumberAtPos(pos: number) {
-        return StringUtils.getLineNumberAtPos(this.getFullText(), pos);
+    getLineAndColumnAtPos(pos: number) {
+        const fullText = this.getFullText();
```
dsherret: Good job using full text.

breaking-changes.md
```diff
@@ -4,6 +4,14 @@ View [CHANGELOG.md](CHANGELOG.md) for more detail on releases. This file is only
 
 ## Version 2
 
+### Removed SourceFile#getLineNumberAtPos()
```
dsherret: Thanks, but could you move this to the bottom? It's not as important to know about as the other items.

src/compiler/ast/module/SourceFile.ts
```diff
@@ -142,11 +142,15 @@ export class SourceFile extends SourceFileBase<ts.SourceFile> {
...
+        return {
+            line: StringUtils.getLineNumberAtPos(fullText, pos),
+            column: StringUtils.getLengthFromLineStartAtPos(fullText, pos)
```
dsherret: I think we should make column 1 indexed?

dsherret:
No, don't worry about changing Node. People can use this method on `SourceFile` to get the column if they want it.

src/compiler/ast/module/SourceFile.ts
cancerberoSgx: I don't know, TypeScript getLineAndCharacterOfPosition() is 0 index based for both things. I was surprised noticing 1-indexed for lines in tsMorph. I think we should be as consistent as possible. But since line numbers are already inconsistent with TS and "column" is a different concept than the TS concept  "Character" maybe is a good idea, so both things are consistent between each other. Should I make the change ?

src/tests/compiler/ast/module/sourceFileTests.ts
cancerberoSgx: IMO there is an error in StringUtils methods: 
```ts
    static getLineNumberAtPos(str: string, pos: number) {
        errors.throwIfOutOfRange(pos, [0, str.length + 1], nameof(pos));
```

should be `errors.throwIfOutOfRange(pos, [0, str.length], nameof(pos));`

right?

But then if I change it, the following test fails, and I think the test is wrong: 

```ts
        it("should not throw if providing a pos the length of the string", () => {
            expect(() => StringUtils.getLineNumberAtPos("", 1)).to.not.throw();
        });
```

Will try to fix this and see if tests are green but maybe you have some idea why you implement this this way? is just an error or is it a hack for something to work?

cancerberoSgx: @dsherret So if I fix StringUtil methods changing the throwIfOutOfRange and change its tests that test explicitly for this (like previous) everything works as expected and tests are green. But I wonder if fixing this issue won't have consequences for user's code ? Is just that, do you want me to push it right here? or perhaps in a separate PR ?

dsherret: Yeah, that's correct and that test is wrong. Should be `0` and not `1`. It can be in this PR. Thanks! :)

cancerberoSgx: Done

src/compiler/ast/module/SourceFile.ts
cancerberoSgx: @dsherret only waiting confirmation on this and I think that would be all for this PR

dsherret: Yeah, the idea of line and column is usually used by editors and for display purposes so keeping it 1-indexed means that people won't have to convert themselves. Monaco editor is 1-indexed I believe (even in the code).

Let's make it 1-indexed and change the description to `Gets the line and column number at the provided position (1-indexed).` Thanks!

cancerberoSgx: on it, you are right, monaco-editor API has the Column and line number concepts and both are  1-indexed

PR #594, opened 2019-03-31
Improve npm test

cancerberoSgx:
* changes "test" script so --invert is not used and I can filter with `npm test -- --grep LanguageService`
 * adds `npm run test-watch` that executes tests whenever a file in `src` changes


package.json
```diff
@@ -12,10 +12,11 @@
     "lint-src-and-fix": "tslint \"src/**/*.ts\" --fix",
     "build": "rimraf dist && npx ttsc && ts-node --transpileOnly scripts/build/removeEmptyFiles",
     "build-declarations": "rimraf dist-declarations && ts-node --max-old-space-size=8192 --transpileOnly scripts/generation/main create-code-block-writer-file create-declaration-file",
-    "test": "cross-env TS_NODE_COMPILER=\"ttypescript\" TS_NODE_TRANSPILE_ONLY=\"true\" mocha --opts mocha.opts --grep @performance --invert",
-    "test-debug": "cross-env TS_NODE_COMPILER=\"ttypescript\" TS_NODE_TRANSPILE_ONLY=\"true\" mocha --opts mocha.opts --grep @performance --invert --inspect-brk",
+    "test": "cross-env TS_NODE_COMPILER=\"ttypescript\" TS_NODE_TRANSPILE_ONLY=\"true\" mocha --opts mocha.opts --grep \"^(?!.*@performance)\"",
+    "test-debug": "cross-env TS_NODE_COMPILER=\"ttypescript\" TS_NODE_TRANSPILE_ONLY=\"true\" mocha --opts mocha.opts --grep \"^(?!.*@performance)\" --inspect-brk",
+    "test-watch": "watch \"npm test\" src",
```
dsherret: There is a mocha `--watch src`. Is using `watch` necessary?

cancerberoSgx: It didn't work for me, but I'm not so familiar with mocha, let me research a little bit.

dsherret: Thanks for looking into that!

Issue #590, opened 2019-03-29
Remove Unused


cancerberoSgx:
Similar to methods `organizeImports()` , `formatText()`, `fixMissingImports()` I would like to have a method `removeUnused()` that removes any variable, parameter, named import, type parameter, method, etc. that is not used. 

```ts
    /**
     *  Removes used declarations like variables, parameters, members, etc. from this node descendants.
     *
     * @param formatSettings - Format code settings.
     * @param userPreferences - User preferences for refactoring.
     */
    removeUnused(formatSettings: FormatCodeSettings = {}, userPreferences: UserPreferences = {}) {}
```

**Alternatives**
The method could accept some options to control which kind of nodes should be removed or not. Also if we decide to put the method in SourceFile instead of Node, it could accept a targetnode option:

```ts
type RemoveUnusedExclude = 'variable'|'parameter'|'typeParameter'|'import'|'function'|'class'|'type'|'method'|'property'|'constructor'|'etc'
interface RemoveUnusedOptions{
    exclude?: RemoveUnusedExclude[]
    targetNode?: Node
}
removeUnused(options?: RemoveUnusedOptions):void
```

**Doubts** 

 * Similarly to formatText() this applies to any node, this method could be declared in `Node` although users will usually call it in `SourceFile`. So I'm not sure where would be the best location..

 * Should we use TS built in code fixes that already implements this? or should we implement it from scratch so Nodes are not forgotten ?

In some cases it could be tricky to define if something is used or not, like type parameters, exports, etc, and the implementation must be 100% safe, but I could try implementing something like that and see...

dsherret:
Would this also include removing functions/classes/interfaces/etc? I think though that for performance reasons we should do it at a project level with the ability to optionally provide a source file or sourcefiles to limit the refactor to. That would allow us to do all the analysis first across the entire project for what to remove and then afterwards do the manipulation. If it's analysis, manipulation, analysis, manipulation, etc. it will be very slow because the compiler will need to build up the symbols each time. (We should also put this in a new `src/compiler/tools/refactors` folder)

Probably it should exist on `LanguageService` and `SourceFile` since that's where the other code fixes are. What about `LanguageService#removeUnusedIdentifiersAndDeclarations(containers?: Node | Node[]): this;`? (it would search all the descendants of the node(s) passed in or all the source files in the project).

dsherret:
Or actually maybe just`removeUnusedDeclarations`.

cancerberoSgx:
[...]

PR #595, opened 2019-03-31
fixUnusedIdentifiers compiler code fix

cancerberoSgx:
https://github.com/dsherret/ts-morph/issues/590

 # Summary

 * using `getCombinedCodeFix()` with fix `unusedIdentifier_delete` similar to `fixMissingImports` implementation
 * tests to verify lots of cases
 * it forgets (some) nodes
 * just a method in SourceFile, no compiler/tools/refactor or any other folder created
 * no support for containerNode because we are using `getCombinedCodeFix()`
 * docs in SourceFile details 

# Doubts
 * Maybe would be better to name the method fixUnusedIdentifier so we name them as TS fix ids (just like fixMissingImports)


src/compiler/ast/module/SourceFile.ts
```diff
@@ -744,6 +744,40 @@ export class SourceFile extends SourceFileBase<ts.SourceFile> {
...
 
+    /**
+     * Removes all unused declarations like interfaces, classes, enums, functions, variables, parameters,
+     * methods, properties, imports, etc. from this file.
+     *
+     * WARNING! This will forget all the nodes in the file! It's best to do this after you're all done with the file.
+     */
+    removeUnusedDeclarations(formatSettings: FormatCodeSettings = {}, userPreferences: UserPreferences = {}) {
+        const sourceFile = this;
+        // heads up: calling it twice removes more unused variables, like things referenced in unused things or named variables
```
dsherret: Doesn't that mean it might need to be called 3 times or 4 or ...etc...? :)

dsherret:
Hmmm... this is not really what I had in mind for this feature and doesn't match what we discussed. For code fixes from the compiler, we should be copying them exactly (side note: don't bother putting so much effort into testing compiler code fixes so much because they are already tested in the compiler for us).

For something like `removeUnusedDeclarations` we could implement it ourselves and limit what nodes are affected by the change, but it would be a bit of work. I think we still need to discuss this more in #590 before we proceed on this. I'll mark those issues as in discussion.

dsherret:
Yes, this would be better named `fixUnusedIdentifiers` to match the code fix. I don't think it's worth it to remember the nodes for this one, so best to just do `this.applyTextChanges` in the body.

I think we could implement this, but it's different from `removeUnusedDeclarations`.

src/compiler/ast/module/SourceFile.ts
cancerberoSgx: Yes, I think so :) Perhaps is better to just call it once? I think is not incorrect not to remove a variable that get's unused after others are removed. I could research a bit more about it.

cancerberoSgx:
> Hmmm... this is not really what I had in mind for this feature and doesn't match what we discussed.

Sure. But this is cheap, safe and will be supported by TypeScript team. I realize we agreed on implementing it from scratch at LanguageService level, but I don't know if I will have time for that scope in the short term. Also I'm not sure if it's worthwhile to work in that while we have this implementation. What do you think? Can we merge this anyway ?

I tried to perform the changes without forgetting nodes but I couldn't. I tried to make a copy of the file and apply the change there, then compare different nodes getting the top-most differences and use removeChildren() manipulation only on those but I couldn't . Do you know if an approach like that is possible? if so would be awesome since we can implement any text-replace without forgetting nodes unless they are modified.

src/compiler/ast/module/SourceFile.ts
```diff
@@ -744,6 +744,40 @@ export class SourceFile extends SourceFileBase<ts.SourceFile> {
         return this;
     }
 
+    /**
+     * Removes all unused declarations like interfaces, classes, enums, functions, variables, parameters,
+     * methods, properties, imports, etc. from this file.
+     *
+     * WARNING! This will forget all the nodes in the file! It's best to do this after you're all done with the file.
+     */
+    removeUnusedDeclarations(formatSettings: FormatCodeSettings = {}, userPreferences: UserPreferences = {}) {
```
dsherret: Rename to `fixUnusedIdentifiers`.

src/compiler/ast/module/SourceFile.ts
dsherret: Yes, it should only be called once.

src/compiler/ast/module/SourceFile.ts
```diff
@@ -744,6 +744,40 @@ export class SourceFile extends SourceFileBase<ts.SourceFile> {
...
+        const sourceFile = this;
+        // heads up: calling it twice removes more unused variables, like things referenced in unused things or named variables
+        applyCombinedCodeFix(this._context.languageService.getCombinedCodeFix(this, "unusedIdentifier_delete", formatSettings, userPreferences));
+        applyCombinedCodeFix(this._context.languageService.getCombinedCodeFix(this, "unusedIdentifier_delete", formatSettings, userPreferences));
```
dsherret: Use:

```ts
this._context.languageService.getCombinedCodeFix(this, "unusedIdentifier_delete", formatSettings, userPreferences).applyChanges();
```

Don't worry about how that will forget all the nodes. We can improve it in the future if necessary.

src/tests/compiler/ast/module/sourceFileTests.ts
```diff
@@ -1300,6 +1300,20 @@ function myFunction(param: MyClass) {
...
+        it("should not forget nodes", () => {
+            const startText = `import { MyClass } from "./MyClass";\n\nconst t = new MyClass();\nconst u: MyInterface = {};\nconst v = new MyClass2();\nconst w = new MyClass3()`;
+            const { sourceFile, project } = getInfoFromText(startText);
+            project.createSourceFile("/MyClass.ts", "export class MyClass {} export class MyClass2 {} export class MyClass3 {}");
+            project.createSourceFile("/MyInterface.ts", "export interface MyInterface {}");
+            const w = sourceFile.getVariableDeclaration("w")!.getFirstDescendantByKindOrThrow(SyntaxKind.Identifier);
+            const i = sourceFile.getImportDeclarations()[0];
+            expect(i.getText()).to.equals("import { MyClass } from \"./MyClass\";");
+            sourceFile.fixMissingImports();
+            expect(() => w.getText()).not.to.throw();
```
dsherret: Check `w.wasForgotten()`. It's a bit more explicit. Some old places might be checking if it throws, but that was before I added `wasForgotten()`.

src/tests/compiler/ast/module/sourceFileTests.ts
```diff
@@ -1405,4 +1419,247 @@ const t = 5;`;
             });
         });
     });
+
+    describe(nameof<SourceFile>(l => l.removeUnusedDeclarations), () => {
+
```
dsherret: See how the spacing is in the other functions. I would use a code formatter on the project, but they sometimes annoy me... I want to implement my own code formatter in the future.

src/tests/compiler/ast/module/sourceFileTests.ts
```diff
@@ -1405,4 +1419,247 @@ const t = 5;`;
...
+        function test(code: string, expected: string) {
```
dsherret: See the name of this function in other tests and copy that.

src/tests/compiler/ast/module/sourceFileTests.ts
```diff
@@ -1405,4 +1419,247 @@ const t = 5;`;
...
+                    import {used, unused} from "bar";
+                    export const c = used + 1;
+                    `, `
+                    import {used} from "bar";
+                    export const c = used + 1;
+                `);
+            });
+        });
+
+        describe(`${nameof<SourceFile>(l => l.removeUnusedDeclarations)} different kinds`, () => {
```
dsherret: Do a single very basic test for this. Not worth adding all these tests.

package.json
```diff
@@ -71,6 +71,7 @@
   "dependencies": {
     "@dsherret/to-absolute-glob": "^2.0.2",
     "code-block-writer": "7.3.0",
+    "conditional-type-checks": "^1.0.0",
```
cancerberoSgx: Since this is is imported in the code itself it should be in dependencies

dsherret: That library is only used at compile time (like nameof).

docs/details/source-files.md
```diff
@@ -248,3 +248,13 @@ To add all the missing import declarations to a source file, use the following m
 ```ts
 sourceFile.fixMissingImports();
 ```
+
+### Remove Unused Identifiers
```
dsherret: Perhaps we should change this to "Fix Unused Identifiers"

docs/manipulation/index.md
```diff
@@ -123,3 +123,7 @@ classDeclaration.getInstanceProperties();
 // you'll need to get the reference again:
 classDeclaration = sourceFile.getClass("MyClass")!;
 ```
+
+## SourceFile Code Fixes
+
+Source files supports a variety of useful code fixes and refactors such as `organizeImports`, `fixMissingImports`, `removeUnusedDeclarations`. Check more details in [source files details page](../details/source-files).
```
dsherret: Thanks for adding this! Just noticed it says `removeUnusedDeclarations` though.

src/compiler/ast/module/SourceFile.ts
```diff
@@ -748,6 +747,22 @@ export class SourceFile extends SourceFileBase<ts.SourceFile> {
         return this;
     }
 
+    /**
+     * Removes all unused declarations like interfaces, classes, enums, functions, variables, parameters,
+     * methods, properties, imports, etc. from this file.
+     *
+     * Tip: Call this method twice in order to obtain optimal results. Sometimes there nodes that are only referenced
```
dsherret: It's good to tell people about how calling this multiple times might be useful, but as pointed out before, saying to call it twice is not necessarily sufficient.

package.json
cancerberoSgx: OK, I'm reverting that change then, but just in case, I'm seeing it being imported in src/compiler/ast/aliases.ts, src/structures/Structure.ts and others at src/structures.

docs/details/source-files.md
```diff
@@ -248,3 +248,13 @@ To add all the missing import declarations to a source file, use the following m
...
+
+To remove all unused declarations from a source file, like variables, classes, members, parameters, type parameters, etc, use the following method:
+
+```ts
+sourceFile.fixUnusedIdentifiers()
```
dsherret: Missing semicolon.

package.json
dsherret: Yeah, that's ok. It goes away at compile time.

dsherret:
Thanks a lot @cancerberoSgx!

Sorry I can be picky, but thanks for working through it :P

ruscon:
@dsherret hi,
is there any way to speed up these functions?

```typescript
sourceFiles.forEach((sourceFile) => {
  sourceFile
    .fixMissingImports()
    .fixUnusedIdentifiers()
    .organizeImports();
});
```

execution takes about a minute in my project

dsherret:
Those are all just calls into the compiler API, IIRC. There’s probably a way it could be sped up, but it might be a lot of work that I don’t have time investigating or doing at the moment unfortunately.

Issue #575, opened 2019-03-22
Leading comments as statements


dsherret:
I'm going to do a PR with a whole proposal for this, but basically the idea would be to make leading comments statements. So calling `#getStatementsWithComments()` on the following code:

```ts
// comment 1
/* comment 2 */
/* comment 3 */ const u; // comment 4
```

...would return these three nodes (obviously not these objects, but nodes like them)...

```ts
{
    kind: SyntaxKind.SingleLineCommentTrivia,
    text: "// comment 1"
}, {
    kind: SyntaxKind.MultiLineCommentTrivia,
    text: "/* comment 2 */"
}, {
    kind: SyntaxKind.VariableStatement,
    // ...etc...
}
```

This would also be done for nodes with members (ex. classes, interfaces, enums, object literal expressions, type element nodes). There will be `#getMembersWithComments()`/`#getPropertiesWithComments()`-like methods.

dsherret:
# Inserting Into Statements Handling Comments

## Prerequisites

* Understand comment ownership: https://github.com/Microsoft/TypeScript/wiki/Architectural-Overview#trivia
* Current api is for inserting statements: `insertStatements(index: number, statements: string | WriterFunction | (string | WriterFunction)[]): Statement[];` where `index` is the compiler API child index of the node.

## Issues

* [#189](https://github.com/dsherret/ts-morph/issues/189) - Node inserted before comment when calling `addClass`-like methods
* [#565](https://github.com/dsherret/ts-morph/issues/565) - Inserting at a specific position

## Overview

The parser rightfully excludes comments from the AST in order to reduce complexity in the compiler. The compiler API then provides a way to get the comments via the `ts.getLeadingCommentRanges` and `ts.getTrailingCommentRanges` functions. These functions are provided the source file text and position to know where to start parsing at.

Although not so useful for the compilation process, comments are useful to know about for programmatic refactoring purposes in certain scenarios.

For example, given the following code:

```ts
// initialize
const u = 5;
```

How could one say to insert after the comment?

```ts
// initialize
const t = 4;
const u = 5;
```

Or before?

```ts
const t = 4;
// this value is special
const u = 5;
```

Right now, there doesn't exist an easy way to do this in ts-morph (possible by inserting with `insertText`, which takes a position).

## Option 1: Insert API Change

One option would be to change the API for `insertStatements` to take a position.

### Upsides

* Close functionality to the compiler API.

### Downsides

* API is not intuitive.
[...]

PR #597, opened 2019-03-31
Implementation of "Extended Comments"

dsherret:
From #575.

Todo:

* [x] Comments returned on classes
* [x] Comments returned on interfaces
* [x] Comments returned on enums
* [x] Comments returned on object literal expressions
* [x] Comments returned on type literal expressions
* [x] Comments returned on case/default clauses
* [x] ~~Way to insert comments on classes/interfaces/etc...~~ Moved to #605
* [x] Check leading/trailing comments for these. Should the behaviour change on other members to not include these extended comments in leading comments? Answer: No, behaviour should remain as unchanged as possible to better mirror the compiler api.
* [x] Change everything to be `.getXWithComments` rather than returning comments for methods like `getStatements()`
* [x] ~~Structures for these comments?~~ No, this won't exist. If anything, the comments should all be included in the leading and trailing trivia.
* [x] ~~Probably make interfaces/classes/etc. structures have a `members` array instead of specific properties... seems like consistency and flexibility might overrule convenience here.~~ No, because there won't be structures for comments. The major motivator for having a `members` array would be maintaining the order of the members which won't be done at the moment.


src/compiler/ast/comment/CompilerComments.ts
```diff
@@ -0,0 +1,120 @@
+import { ts, SyntaxKind } from "../../../typescript";
+
+export class CompilerExtendedComment implements ts.Node {
```
dsherret: Add tests for this.

src/manipulation/nodeHandlers/DefaultParentHandler.ts
```diff
@@ -34,41 +33,40 @@ export class DefaultParentHandler implements NodeHandler {
...
         let count = this.childCount;
 
         // handle any custom mappings
         this.handleCustomMappings(newNode);
 
         // get the first child
-        while (!currentNodeChildren.done && !newNodeChildren.done && !this.isFirstChild(currentNodeChildren.peek, newNodeChildren.peek))
-            this.helper.handleForValues(this.straightReplacementNodeHandler, currentNodeChildren.next(), newNodeChildren.next(), newSourceFile);
+        while (!currentChildren.done && !newChildren.done && !this.isFirstChild(currentChildren.peek, newChildren.peek))
+            this.helper.handleForValues(this.straightReplacementNodeHandler, currentChildren.next(), newChildren.next(), newSourceFile);
```
dsherret: Add a few tests that ensure the comments stick around between manipulations.

src/tests/compiler/ast/utils/extendedCommentParserTests.ts
```diff
@@ -0,0 +1,273 @@
...
+
+        it("should not get comments on the same line", () => {
+            doTest("class c {\n/*a*/p;/*b*/ //c\n}", [{
+                kind: ts.SyntaxKind.PropertyDeclaration,
+                pos: 9,
+                end: 17
+            }]);
+        });
+
+        it("should ignore comments that are on the opening brace", () => {
```
dsherret: I think comments before the closing brace should be parsed. For example:

```ts
class C {
    prop: string;
/* test */ }
```

Even though it's on the same line as the close brace, it feels like a comment statement.

src/utils/TypeGuards.ts
```diff
@@ -49,6 +49,54 @@ export class TypeGuards {
...
+    /**
+     * Gets if the provided value is a Node.
+     */
+    static isNode(value: unknown): value is compiler.Node {
+        if (value == null)
+            return false;
+        const kind = (value as any)["kind"];
+        if (typeof kind !== "number")
+            return false;
+        return kind !== SyntaxKind.SingleLineCommentTrivia
```
dsherret: Not necessarily true anymore. This is wrong.

dsherret: Heh. This whole method is wrong... it's checking if the passed in object is actually a ts.Node

dsherret:
Will do one final code review for this on a different day, merge this, then publish a release candidate.

src/utils/ArrayUtils.ts
```diff
@@ -37,6 +37,7 @@ export class ArrayUtils {
         return items.reduce((a, b) => a.concat(b), []);
     }
 
+    // todo: remove and use Array.prototype.find now that es5 support is dropped
```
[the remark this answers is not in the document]
dsherret: Will do after merge.

PR #601, opened 2019-04-03
RefactorEditInfo#applyChanges()

cancerberoSgx:
* fix issue in getTextFromFormattingEdits that was breaking "move to a new file" refactor. 
 * adds a test in fileTextChangesTest that reproduces previous one
 * adds RefactorEditInfo#applyChanges()
 * adds tests for previous one
 * fix error in applyChanges#applyChanges()

### DIFF
--- a/src/compiler/tools/results/FileTextChanges.ts
+++ b/src/compiler/tools/results/FileTextChanges.ts
@@ -6,7 +6,9 @@ import { Memoize } from "../../../utils";
 import { TextChange } from "./TextChange";
 
 export interface ApplyFileTextChangesOptions {
-    /** If a file should be overwritten when the file text change is for a new file, but the file currently exists. */
+    /**
+     * If a file should be overwritten when the file text change is for a new file, but the file currently exists.
+     */
     overwrite?: boolean;
 }
 
--- a/src/compiler/tools/results/FileTextChanges.ts
+++ b/src/compiler/tools/results/FileTextChanges.ts
@@ -59,7 +59,7 @@ export class FileTextChanges {
         if (this._isApplied)
             return;
 
-        let file: SourceFile | undefined;
+        let file = this.getSourceFile();
 
         if (this.isNewFile() && file != null && !options.overwrite) {
             throw new errors.InvalidOperationError(`Cannot apply file text change for creating a new file when the ` +
@@ -68,8 +68,6 @@ export class FileTextChanges {
 
         if (this.isNewFile())
             file = this._context.project.createSourceFile(this.getFilePath(), "", { overwrite: options.overwrite });
-        else
-            file = this.getSourceFile();
 
         if (file == null) {
             throw new errors.InvalidOperationError(`Cannot apply file text change to modify existing file ` +
--- a/src/manipulation/formatting/getTextFromFormattingEdits.ts
+++ b/src/manipulation/formatting/getTextFromFormattingEdits.ts
@@ -1,14 +1,7 @@
 ﻿import { SourceFile, TextChange } from "../../compiler";
 
 export function getTextFromFormattingEdits(sourceFile: SourceFile, formattingEdits: ReadonlyArray<TextChange>) {
-    // reverse the order
-    formattingEdits = [...formattingEdits].sort((a, b) => {
-        const aStart = a.getSpan().getStart();
-        const bStart = b.getSpan().getStart();
-        const difference = bStart - aStart;
-
-        return difference === 0 ? 1 : difference; // reverse when equal
-    });
+    formattingEdits = [...formattingEdits].reverse();
     let text = sourceFile.getFullText();
 
     for (const textChange of formattingEdits) {
--- a/src/manipulation/formatting/getTextFromFormattingEdits.ts
+++ b/src/manipulation/formatting/getTextFromFormattingEdits.ts
@@ -1,13 +1,17 @@
 ﻿import { SourceFile, TextChange } from "../../compiler";
 
 export function getTextFromFormattingEdits(sourceFile: SourceFile, formattingEdits: ReadonlyArray<TextChange>) {
-    formattingEdits = [...formattingEdits].reverse();
+    // reverse the order
+    formattingEdits = [...formattingEdits].sort((a, b) => {
+        const aStart = a.getSpan().getStart();
+        const bStart = b.getSpan().getStart();
+        const difference = bStart - aStart;
+        return difference === 0 ? -1 : difference < 0 ? -1 : 1;
+    });
     let text = sourceFile.getFullText();
-
     for (const textChange of formattingEdits) {
         const span = textChange.getSpan();
         text = text.slice(0, span.getStart()) + textChange.getNewText() + text.slice(span.getEnd());
     }
-
     return text;
 }
--- a/src/tests/compiler/tools/results/fileTextChangesTests.ts
+++ b/src/tests/compiler/tools/results/fileTextChangesTests.ts
@@ -64,5 +64,17 @@ describe(nameof(FileTextChanges), () => {
             change.applyChanges(); // should do nothing if called twice
             expect(change.getSourceFile()!.getText()).to.equal("text; const t; const u;");
         });
+
+        it("should apply file text change that first inserts and then removes ('move to a new file' refactor edits changes)", () => {
+          const project = setup();
+          const change = new FileTextChanges(project._context, {
+              fileName: "test.ts",
+              isNewFile: false,
+              textChanges: [{ newText: "text; ", span: { start: 0, length: 0 } }, {newText: "", span: {start: 0, length: 9}}]
+          });
+          change.applyChanges();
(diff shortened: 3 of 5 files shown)


src/compiler/tools/results/FileTextChanges.ts
```diff
@@ -59,7 +61,7 @@ export class FileTextChanges {
         if (this._isApplied)
             return;
 
-        let file: SourceFile | undefined;
+        let file = this.getSourceFile();
 
         if (this.isNewFile() && file != null && !options.overwrite) {
```
cancerberoSgx: this expression was always false

src/manipulation/formatting/getTextFromFormattingEdits.ts
```diff
@@ -1,14 +1,7 @@
...
 export function getTextFromFormattingEdits(sourceFile: SourceFile, formattingEdits: ReadonlyArray<TextChange>) {
-    // reverse the order
-    formattingEdits = [...formattingEdits].sort((a, b) => {
-        const aStart = a.getSpan().getStart();
-        const bStart = b.getSpan().getStart();
-        const difference = bStart - aStart;
-
-        return difference === 0 ? 1 : difference; // reverse when equal
-    });
+    formattingEdits = [...formattingEdits].reverse();
```
cancerberoSgx: the problem here was that the array was not reversed in case two items have the same start, this was breaking "move to a new file" refactor. Check fileTestChangesTest

src/tests/compiler/tools/results/fileTextChangesTests.ts
```diff
@@ -64,5 +64,17 @@ describe(nameof(FileTextChanges), () => {
             change.applyChanges(); // should do nothing if called twice
             expect(change.getSourceFile()!.getText()).to.equal("text; const t; const u;");
         });
+
+        it("should apply file text change that first inserts and then removes ('move to a new file' refactor edits changes)", () => {
```
cancerberoSgx: this wasn't working

src/compiler/tools/results/FileTextChanges.ts
dsherret: Thanks! I recently broke this.

src/manipulation/formatting/getTextFromFormattingEdits.ts
dsherret: Hmmm... this no longer sorts though. Could you change it back to sort, but still fix the issue?

src/compiler/tools/results/FileTextChanges.ts
dsherret: Is there a test for this? We should add one if not.

cancerberoSgx: Yes I think there's already tests to verify "should throw if a file text change instruct to create a new file that already exists" . (and overwrite is not given)

src/manipulation/formatting/getTextFromFormattingEdits.ts
cancerberoSgx: done

dsherret:
Hmmm... I didn't notice the failing tests. I'll fix this up.

src/manipulation/formatting/getTextFromFormattingEdits.ts
```diff
@@ -6,15 +6,12 @@ export function getTextFromFormattingEdits(sourceFile: SourceFile, formattingEdi
         const aStart = a.getSpan().getStart();
         const bStart = b.getSpan().getStart();
         const difference = bStart - aStart;
-
-        return difference === 0 ? 1 : difference; // reverse when equal
+        return difference === 0 ? -1 : difference < 0 ? -1 : 1;
```
dsherret: @cancerberoSgx I just had to revert this line to fix the tests. Do you have a test that requires it being what you set it to here?

src/compiler/tools/results/FileTextChanges.ts
dsherret: I just tried reverting the code here and the tests pass. I'll add a test for this.

src/manipulation/formatting/getTextFromFormattingEdits.ts
cancerberoSgx: Mmm, the file text changes that involve multiple files and spans with same starts are broken on my setup. See below. I'm now making 100% sure that's not something wrong with my local - downloading your dev branch from scratch. 

Just in case I?m using node 11.6 - the problem was the same as before, reversing two spans with equal starts break refactor text changes when they first remove and then add (that was my conclusion while debugging the problem and that's why I change this line. ) But before saying more let me check with a fresh branch, and confirm. Question, why reversing the changes here ? performance ? forgetting less nodes ? 

Well let me check and come back to you, Thanks! (BTW the tests were green here and now red so something is not right... )

npm test errors:

```
1) FileTextChanges
       applyChanges
         should apply file text change that first inserts and then removes ('move to a new file' refactor edits changes):

      AssertionError: expected 'st t; const u;' to equal 'text; const u;'
      + expected - actual

      -st t; const u;
      +text; const u;

      at Context.it (src/tests/compiler/tools/results/fileTextChangesTests.ts:81:56)

  2) RefactorEditInfo
       applyChanges
         should apply the refactor 'Move to a new file':

      AssertionError: expected ') { return `${b} ${a}`; }\nexport const c = z(1, \'a\')' to equal 'import { z } from "./z";\n\nexport const c = z(1, \'a\')'
      + expected - actual

      -) { return `${b} ${a}`; }
      +import { z } from "./z";
      +
       export const c = z(1, 'a')

      at Context.it (src/tests/compiler/tools/results/refactorEditInfoTests.ts:15:49)

```

dsherret: That's very strange! Thanks for looking into it.

It's reversed because if you manipulate a string going from back to front you don't have to keep an offset count.

cancerberoSgx: Update: I tried from scratch, cloning this project, checking out dev branch and "yarn test" will fail with the same errors as above. 

Also a couple of notes regarding the install. 

"yarn test" thrown 

```
yarn run v1.15.2
warning package.json: "dependencies" has dependency "typescript" with range "^3.4.1" that collides with a dependency in "devDependencies" of the same name with version "3.4.1"
$ cross-env TS_NODE_COMPILER="ttypescript" TS_NODE_TRANSPILE_ONLY="true" mocha --opts mocha.opts --grep "^(?!.*@performance)"
Error: ENOENT: no such file or directory, open '/Users/sebastiangurin/git/ts-morph/node_modules/typescript-3.4.2/lib/lib.d.ts'
    at Object.openSync (fs.js:450:3)
    at Object.readFileSync (fs.js:350:35)
....
```

So I did the following to "fix it":

```
 cp -r node_modules/typescript-3.4.1/ node_modules/typescript3.4.2
```

That's the only "hack" I needed to perform in order the project to compile and run the tests. Do you think this can be the cause of discrepancies ?

Until now I was working with npm,  and for it to work I had to remove the dependency aliases for typescript from package.json. I thought that could be the reason, but no I tried the installation 100% with yarn and same results...

cancerberoSgx: Also regarding the algorithm, I was investigating a little bit inside TypeScript and seems they do the same, with a small detail: They put the shorter span first in case two of them start in the same position: https://github.com/Microsoft/TypeScript/blob/ef4acc8841676064ac3c29c8d6b5d58fe9a30e3c/src/services/textChanges.ts#L774

Will try to implement that in ts-morph to see if that fix my tests. 

Although I'm concerned why you are not getting the same errors as me.

cancerberoSgx: This is working : 

```
    // reverse the order
    formattingEdits = [...formattingEdits].sort((a, b) => {
        const aStart = a.getSpan().getStart();
        const bStart = b.getSpan().getStart();
        const difference = bStart - aStart;
        // If the start position is the same, put the shorter range first, since an empty range (x, x) may precede (x, y) but not vice-versa.
        return difference === 0 ? (a.getSpan().getLength() > b.getSpan().getLength() ? -1 : 1) : difference;
    });
```

@dsherret  Should I create a PR for this ?

dsherret: That's very strange. Maybe try deleting `node_modules` and rerun `yarn install`. It should have been using 3.4.1 so that's strange. (I just upgraded the repo to 3.4.3 now). That copy fix you have to do might be a clue... perhaps a tool isn't doing something correctly.

Regarding in the compiler, that seems to be `Changes` though and not `TextChanges`? I would assume what we receive from the compiler should always be in the order that the changes should be made in and if not then that would be a bug in the compiler.

It's strange that it works for me and on the CI, but not on your computer... I think we need to figure out why... it may just be a dependency issue and reinstalling the node_modules with yarn might help.

Issue #627, opened 2019-05-24
Support custom module resolution


dsherret:
From #626.

dsherret:
Implemented in #628.

PR #628, opened 2019-05-24
Custom module and type reference directive resolution

DanTsk:
For #627

### DIFF
--- a/lib/ts-morph.d.ts
+++ b/lib/ts-morph.d.ts
@@ -361,6 +361,10 @@ export interface ProjectOptions {
      * @remarks Consider using `useVirtualFileSystem` instead.
      */
     fileSystem?: FileSystemHost;
+    /**
+     * Custom overrides for Compiler Host that don't overlap with sys default
+     */
+    resolutionHost?: ResolutionHostFactory;
 }
 
 /**
@@ -3414,7 +3418,7 @@ export declare class BindingElement extends BindingElementBase<ts.BindingElement
      *
      * For example in `const { a: b } = { a: 5 }`, `a` would be the property name.
      */
-    getPropertyNameNode(): NumericLiteral | StringLiteral | Identifier | ComputedPropertyName | undefined;
+    getPropertyNameNode(): Identifier | NumericLiteral | StringLiteral | ComputedPropertyName | undefined;
 }
 
 export declare class ObjectBindingPattern extends Node<ts.ObjectBindingPattern> {
@@ -6453,7 +6457,7 @@ export declare class JsxText extends JsxTextBase<ts.JsxText> {
     /**
      * Gets if the JSX text contains only white spaces.
      */
-    containsOnlyTriviaWhiteSpaces(): boolean;
+    containsOnlyTriviaWhiteSpaces(): any;
 }
 
 export interface ImplementedKindToNodeMappings {
@@ -8914,7 +8918,7 @@ export declare class Signature {
     /**
      * Gets the signature's declaration.
      */
-    getDeclaration(): MethodSignature | MethodDeclaration | ConstructorDeclaration | GetAccessorDeclaration | SetAccessorDeclaration | CallSignatureDeclaration | ConstructSignatureDeclaration | IndexSignatureDeclaration | FunctionTypeNode | ConstructorTypeNode | FunctionExpression | ArrowFunction | FunctionDeclaration | JSDocFunctionType;
+    getDeclaration(): ArrowFunction | MethodSignature | MethodDeclaration | ConstructorDeclaration | GetAccessorDeclaration | SetAccessorDeclaration | CallSignatureDeclaration | ConstructSignatureDeclaration | IndexSignatureDeclaration | FunctionTypeNode | ConstructorTypeNode | FunctionExpression | FunctionDeclaration | JSDocFunctionType;
 }
 
 export declare class Symbol {
@@ -9046,6 +9050,14 @@ export interface RenameOptions {
  */
 export interface UserPreferences extends ts.UserPreferences {
 }
+
+export interface ResolutionHostOverrides {
+    resolveModuleNames?(moduleNames: string[], containingFile: string, reusedNames?: string[], ref?: ts.ResolvedProjectReference): (ts.ResolvedModule | undefined)[];
+    getResolvedModuleWithFailedLookupLocationsFromCache?(modulename: string, containingFile: string): ts.ResolvedModuleWithFailedLookupLocations | undefined;
+    resolveTypeReferenceDirectives?(typeDirectiveNames: string[], containingFile: string, ref?: ts.ResolvedProjectReference): (ts.ResolvedTypeReferenceDirective | undefined)[];
+}
+
+export declare type ResolutionHostFactory = (fileSystem: FileSystemHost) => ResolutionHostOverrides;
 
 export declare class LanguageService {
     private readonly _compilerObject;
--- a/src/Project.ts
+++ b/src/Project.ts
@@ -1,5 +1,5 @@
 import { CodeBlockWriter } from "./codeBlockWriter";
-import { Diagnostic, EmitOptions, EmitResult, LanguageService, Node, Program, SourceFile, TypeChecker, CompilerHostFactory } from "./compiler";
+import { Diagnostic, EmitOptions, EmitResult, LanguageService, Node, Program, SourceFile, TypeChecker, ResolutionHostFactory } from "./compiler";
 import * as errors from "./errors";
 import { DefaultFileSystemHost, Directory, DirectoryAddOptions, FileSystemHost, FileSystemWrapper, VirtualFileSystemHost } from "./fileSystem";
 import { ProjectContext } from "./ProjectContext";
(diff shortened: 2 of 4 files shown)


src/compiler/tools/LanguageService.ts
```diff
@@ -11,6 +11,18 @@ import { Program } from "./Program";
 import { DefinitionInfo, EmitOutput, FileTextChanges, ImplementationLocation, RenameLocation, TextChange, DiagnosticWithLocation, RefactorEditInfo, CodeFixAction,
     CombinedCodeActions } from "./results";
 
+export interface CompilerHostOverrides {
+    resolveModuleNames?(moduleNames: string[], containingFile: string, reusedNames?: string[], ref?: ts.ResolvedProjectReference): (ts.ResolvedModule | undefined)[];
+    getResolvedModuleWithFailedLookupLocationsFromCache?(modulename: string, containingFile: string): ts.ResolvedModuleWithFailedLookupLocations | undefined;
+    resolveTypeReferenceDirectives?(typeDirectiveNames: string[], containingFile: string, ref?: ts.ResolvedProjectReference): (ts.ResolvedTypeReferenceDirective | undefined)[];
+}
+
+export type CompilerHostFactory = (languageService: ts.LanguageServiceHost) => CompilerHostOverrides;
```
dsherret: Thanks! Will look at this PR later. Question though: why does it require a language service host? Could you provide an example of some code that uses this?

DanTsk: Sure, everything is pretty clear. First of all,  `LanguageService` implements almost everything that is the need for work with sources and file system, so it is really useful for module resolution.

And another thing is `resolveTypeReferenceDirectives` implementation.  The default implementation is referenced from `ts.resolveTypeReferenceDirective` function, that is exposed directly on `ts` namespace. This function has the following signature: 
```
function resolveTypeReferenceDirective(typeReferenceDirectiveName: string, containingFile: string | undefined, options: CompilerOptions, host: ModuleResolutionHost, redirectedReference?: ResolvedProjectReference): ResolvedTypeReferenceDirectiveWithFailedLookupLocations;
```
As you can see, `ModuleResolutionHost` is required argument that can be build only with 'LanguageService' implemenatations. Here is `ModuleResolutionHost` interface:
```
 interface ModuleResolutionHost {
        fileExists(fileName: string): boolean;
        readFile(fileName: string): string | undefined;
        trace?(s: string): void;
        directoryExists?(directoryName: string): boolean;    
        realpath?(path: string): string;
        getCurrentDirectory?(): string;
        getDirectories?(path: string): string[];
    }
```
Everything about FS. 
Such a strategy is used in TypeScript - (source)[https://github.com/microsoft/TypeScript/blob/master/src/harness/harnessLanguageService.ts#L295] where `NativeLanguageServiceHost` is an implementation of `ts.LanguageServiceHost`

dsherret: So it seems like the main thing that's absolutely necessary is the ability to have access to the project's file system?

I'm thinking that this method should instead be passed a `FileSystemHost` and then someone could construct a `ts.ModuleResolutionHost` from that if they need it (and perhaps a helper function could be added so people could easily create a module resolution host from a `FileSystemHost`). I'd rather not expose the internally used `ts.CompilerHost` because then it becomes part of the public api.

DanTsk: For example, we can inject `FileSystemHost` from `Project`.

src/compiler/tools/LanguageService.ts
```diff
@@ -82,7 +94,11 @@ export class LanguageService {
             useCaseSensitiveFileNames: () => languageServiceHost.useCaseSensitiveFileNames!(),
             getNewLine: () => languageServiceHost.getNewLine!(),
             getEnvironmentVariable: (name: string) => process.env[name],
-            directoryExists: dirName => languageServiceHost.directoryExists!(dirName)
+            directoryExists: dirName => languageServiceHost.directoryExists!(dirName),
+
+            // Apply implementations safely to avoid overrides for predefined
+            resolveModuleNames: compilerHost.resolveModuleNames,
+            resolveTypeReferenceDirectives: compilerHost.resolveTypeReferenceDirectives
```
dsherret: I think we'll also need to add these to `languageServiceHost` above.

src/compiler/tools/LanguageService.ts
```diff
@@ -11,6 +11,18 @@ import { Program } from "./Program";
...
 
+export interface CompilerHostOverrides {
+    resolveModuleNames?(moduleNames: string[], containingFile: string, reusedNames?: string[], ref?: ts.ResolvedProjectReference): (ts.ResolvedModule | undefined)[];
+    getResolvedModuleWithFailedLookupLocationsFromCache?(modulename: string, containingFile: string): ts.ResolvedModuleWithFailedLookupLocations | undefined;
+    resolveTypeReferenceDirectives?(typeDirectiveNames: string[], containingFile: string, ref?: ts.ResolvedProjectReference): (ts.ResolvedTypeReferenceDirective | undefined)[];
+}
+
+export type CompilerHostFactory = (languageService: FileSystemHost) => CompilerHostOverrides;
+
+export interface LanguageServiceOptions {
```
dsherret: We could mark this with `/** @internal */` since the constructor is private in the declaration file.

src/Project.ts
```diff
@@ -59,8 +61,11 @@ export class Project {
         const tsConfigResolver = options.tsConfigFilePath == null ? undefined : new TsConfigResolver(fileSystemWrapper, options.tsConfigFilePath, getEncoding());
         const compilerOptions = getCompilerOptions();
 
+        // init compiler host overrides
+        const compilerHost = options.compilerHostFactoty ? options.compilerHostFactoty(fileSystem) : {};
```
dsherret: The best part about TypeScript is how it helps developers consistently misspell a word ;) (Re: `Factoty`)

src/compiler/tools/LanguageService.ts
dsherret: When you're done with all these changes, run the command `yarn overwrite-declaration-files`. It will update the declaration file stored in the repo (then check that in). Thanks!

dsherret:
By the way, I was trying to think of what a user approaching the library who had never used the compiler API would expect if they wanted to override module resolution. The thing is, it's not a common thing to do so I think just passing these directly to the compiler API like what you're doing here is fine. Thanks again!

lib/ts-morph.d.ts
```diff
@@ -6453,7 +6457,7 @@ export declare class JsxText extends JsxTextBase<ts.JsxText> {
     /**
      * Gets if the JSX text contains only white spaces.
      */
-    containsOnlyTriviaWhiteSpaces(): boolean;
+    containsOnlyTriviaWhiteSpaces(): any;
```
dsherret: Hmmm... I wonder why this happened. Did you use `yarn install` to install the dependencies or `npm install`?

src/compiler/tools/LanguageService.ts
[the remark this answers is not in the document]
dsherret: Ah, ok I see that's not on `ts.CompilerHost`.

lib/ts-morph.d.ts
```diff
@@ -361,6 +361,10 @@ export interface ProjectOptions {
      * @remarks Consider using `useVirtualFileSystem` instead.
      */
     fileSystem?: FileSystemHost;
+    /**
+     * Custom overrides for Compiler Host that don't overlap with sys default
```
dsherret: Could probably use a new comment now that this is renamed.

DanTsk:
@dsherret Thanks for your reviews! Sorry for that misspellings, I was really sleepy😄 . It would be great if you can add some tests, as I am not familiar with the styles of testing in the project. (Noticed that `nameof` is used, pretty interesting approach). 
And us it possible to have a version patch with this functionality published on NPM in the next days?

dsherret:
@DanTsk yeah, no problem! I'll add some tests and then publish sometime today or tomorrow.

Using nameof is very nice because when I refactor the method names the test descriptions update automatically. It's using [this project](https://github.com/dsherret/ts-nameof).

By the way, I added the `createModuleResolutionHost` helper function:

https://github.com/dsherret/ts-morph/commit/3c89a0bd043b4bb19ba2b79b83eb7e160051d20a

dsherret:
Thanks again! I'll do a release after adding some tests.

dsherret:
Oh wow, I'm so glad I tested this. I really screwed up and forgot that files created in the project won't even be in the file system host until they're saved. My bad!

I'm going to change this to provide a module resolution host and that host will inspect the current state of the project. I'm also going to revert my change in https://github.com/dsherret/ts-morph/commit/3c89a0bd043b4bb19ba2b79b83eb7e160051d20a and instead add a new method on project that returns one of those.

dsherret:
Ok, this is released now in v2.2.

See https://dsherret.github.io/ts-morph/setup/index#custom-module-resolution-advanced for an example of setting it up.

DanTsk:
Wow, great! Thank you so much!

PR #635, opened 2019-05-29
Add return statement writer function

marikaner:
Adds one more writer function for return statements.

### DIFF
--- a/src/structurePrinters/WriterFunctions.ts
+++ b/src/structurePrinters/WriterFunctions.ts
@@ -82,8 +82,8 @@ export class WriterFunctions {
      */
     static returnStatement(value: WriterFunctionOrValue): WriterFunction {
         return (writer: CodeBlockWriter) => {
-            writeSeparatedByString(writer, ' ', ['return', value]);
-            writer.write(';');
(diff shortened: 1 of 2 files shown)


dsherret:
@marikaner thanks a lot for the PR. Looks great! I'll merge this tonight after I get home from work.

----

You don't have to bother implementing this (because an easy way of doing it currently isn't in this version... I have a branch with it), but I'm thinking this should print with hanging indentation. So if someone does a newline in the return statement expression then it should print this:

```ts
return someCondition
    && otherCondition
    && finalCondition;
```

It would probably be nice though if it didn't use hanging indentation when inside braces though, but I'd implement that later:

```ts
// looks better IMO
return {
    prop
};

// probably will just print this for now to make it easy
return {
        prop
    };
```

Thoughts?

marikaner:
I agree and I agree that those two cases should be handled differently in the best case. Thanks!

dsherret:
@marikaner thanks so much once again! This is released now in 2.3.0.

I will do the hanging indent change when releasing 3.0. For now it doesn't do any hanging indentation.

PR #667, opened 2019-07-26
Fix typo word: "dependent"

tw3:


dsherret:
Thanks so much for this, @tw3! Really appreciate this.

Issue #669, opened 2019-07-30
add second parameter to getParentWhile condition


Validark:
Builds off: https://github.com/dsherret/ts-morph/issues/107

Being able to call `node.getParentWhile(p => ts.TypeGuards.etc(p) || etc)` is very nice. However, I would like to be able to easily check whether the node under it has the proper relationship to the current parent. For example, let's say I want to figure out whether something is in the conditional of an if statement:

```ts
node.getParentWhile((p, n) => ts.TypeGuards.isIfStatement(p) && p.getExpression() === n)
```

Where:
-  `n` is the current node which starts as `node` in the above example
-  `p` is the current parent which starts as `node.getParent()` in the above example

dsherret:
@Validark that sounds like a good idea. Would you like to implement it? If not, I'll get to it sometime soon.

PR #670, opened 2019-07-30
Implement #669

Validark:
Closes https://github.com/dsherret/ts-morph/issues/669

The whitespace might need fixing.

### DIFF
--- a/src/compiler/ast/common/Node.ts
+++ b/src/compiler/ast/common/Node.ts
@@ -1108,14 +1108,14 @@ export class Node<NodeType extends ts.Node = ts.Node> {
      * Throws if the initial parent doesn't match the condition.
      * @param condition - Condition that tests the parent to see if the expression is true.
      */
-    getParentWhileOrThrow<T extends Node>(condition: (node: Node) => node is T): T;
+    getParentWhileOrThrow<T extends Node>(condition: (parent: Node, node: Node) => parent is T): T;
     /**
      * Goes up the parents (ancestors) of the node while a condition is true.
      * Throws if the initial parent doesn't match the condition.
      * @param condition - Condition that tests the parent to see if the expression is true.
      */
-    getParentWhileOrThrow(condition: (node: Node) => boolean): Node;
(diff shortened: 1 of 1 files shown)


src/compiler/ast/common/Node.ts
```diff
@@ -1124,20 +1124,22 @@ export class Node<NodeType extends ts.Node = ts.Node> {
...
-    getParentWhile(condition: (node: Node) => boolean) {
-        let node: Node | undefined = undefined;
-        let nextParent: Node | undefined = this.getParent();
-        while (nextParent != null && condition(nextParent)) {
-            node = nextParent;
-            nextParent = nextParent.getParent();
-        }
+    getParentWhile(condition: (parent: Node, node: Node) => boolean): Node | undefined;
+    getParentWhile(condition: (parent: Node, node: Node) => boolean) {
+        let node: Node | undefined;
```
dsherret: Bit of a nitpick, but I think you could assign `this` to `node` here then keep using a `while` loop?

src/compiler/ast/common/Node.ts
```diff
@@ -1124,20 +1124,22 @@ export class Node<NodeType extends ts.Node = ts.Node> {
...
-    getParentWhile(condition: (node: Node) => boolean): Node | undefined;
-    getParentWhile(condition: (node: Node) => boolean) {
-        let node: Node | undefined = undefined;
-        let nextParent: Node | undefined = this.getParent();
-        while (nextParent != null && condition(nextParent)) {
-            node = nextParent;
-            nextParent = nextParent.getParent();
-        }
+    getParentWhile(condition: (parent: Node, node: Node) => boolean): Node | undefined;
+    getParentWhile(condition: (parent: Node, node: Node) => boolean) {
```
dsherret: Perhaps we should name `node` as `child`? That way someone won't think this was the original node. Or perhaps there's something better?

dsherret:
Thanks @Validark! This looks good! I'll merge it, but was wondering if you'd be willing to add a test for this new parameter to ensure it won't accidentally break in the future?

PR #699, opened 2019-09-07
fix typo in /docs/manipulation/index.md

tomByrer:
I think you meant `await directory.copyImmediately("CopiedDir");` is coming soon?

### DIFF
--- a/docs/manipulation/index.md
+++ b/docs/manipulation/index.md
@@ -33,8 +33,8 @@ await sourceFile.moveImmediately("movedFile.ts");
 await directory.save();
 await directory.deleteImmediately();
 // todo: these are coming soon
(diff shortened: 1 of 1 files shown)


dsherret:
Hey @tomByrer, yup it looks like I originally definitely meant to say that, but it also looks like `copyImmediately` is implemented on `Directory` :)

docs/manipulation/index.md
```diff
@@ -33,8 +33,8 @@ await sourceFile.moveImmediately("movedFile.ts");
 await directory.save();
 await directory.deleteImmediately();
 // todo: these are coming soon
```
dsherret: Seems like we can remove the todo as well!

PR #703, opened 2019-09-18
feat: add TypeGuards.is

lazarljubenovic:
So I've had this idea that `TypeGuards` should have a generic `is` (name subject to change, of course), which most of other type guards would use. This is similar to how I like using a curried [`isInstanceOf`](https://github.com/lazarljubenovic/type-guards/blob/8eaecf9c00bb7f63c7ea86f24da2bb3b9c910e4f/src/index.ts#L21-L26) utility: you can easily create things like `const isDate = isInstanceOf(Date)`.

I've quickly skimmed through the newly generated output, and it seems like it was basically a transformation from

```ts
    /**
     * Gets if the node is an AnyKeyword.
     * @param node - Node to check.
     */
    static isAnyKeyword(node: compiler.Node): node is compiler.Expression {
        return node.getKind() === SyntaxKind.AnyKeyword;
    }
```

to

```ts
    /**
     * Gets if the node is an AnyKeyword.
     */
    static isAnyKeyword = TypeGuards.is(SyntaxKind.AnyKeyword);
```

These changes are made only for those syntax kinds listed in `ImplementedKindToNodeMappings` (in `src/compiler/ast/kindToNodeMappings.ts`).

This should allow us better composition for methods which accept `kind` and somehow filter out the result based on the kind (and affect the return type).

# Issue

However, `isJSDocTag` was previously this.

```ts
    /**
     * Gets if the node is a JSDocTag.
     * @param node - Node to check.
     */
    static isJSDocTag(node: compiler.Node): node is compiler.JSDocTag {
        switch (node.getKind()) {
            case SyntaxKind.JSDocAugmentsTag:
            case SyntaxKind.JSDocClassTag:
            case SyntaxKind.JSDocParameterTag:
            case SyntaxKind.JSDocPropertyTag:
            case SyntaxKind.JSDocReturnTag:
            case SyntaxKind.JSDocTypedefTag:
            case SyntaxKind.JSDocTypeTag:
            case SyntaxKind.JSDocTag:
                return true;
            default:
                return false;
        }
    }
```

I'm confused since this is _the only_ one (at least according to tests) which doesn't follow the pattern. And, indeed, doing `TypeGuards.isJSDocTag(x)` isn't the same as doing `x.getKind() == SyntaxKind.isJSDocTag`, which is true for all other `ImplementedKindToNodeMappings`. At least I think so -- the diff for `TypeGuards.ts` isn't the nicest because I couldn't figure out how to sort it the same way), so there's no quick way for me check that I can think of.

Maybe my conclusion was wrong, but hopefully this will at least serve as a nice starting point (or a proof that this isn't possible).

### DIFF
--- a/lib/ts-morph.d.ts
+++ b/lib/ts-morph.d.ts
@@ -730,6 +730,530 @@ export declare function getCompilerOptionsFromTsConfig(filePath: string, options
  * Type guards for checking the type of a node.
  */
 export declare class TypeGuards {
+    /**
+     * Gets if the node is an AnyKeyword.
+     */
+    static isAnyKeyword: (node: Node<Node>) => node is Expression<Expression>;
+    /**
+     * Gets if the node is an ArrayBindingPattern.
+     */
+    static isArrayBindingPattern: (node: Node<Node>) => node is ArrayBindingPattern;
+    /**
+     * Gets if the node is an ArrayLiteralExpression.
+     */
+    static isArrayLiteralExpression: (node: Node<Node>) => node is ArrayLiteralExpression;
+    /**
+     * Gets if the node is an ArrowFunction.
+     */
+    static isArrowFunction: (node: Node<Node>) => node is ArrowFunction;
+    /**
+     * Gets if the node is an AsExpression.
+     */
+    static isAsExpression: (node: Node<Node>) => node is AsExpression;
+    /**
+     * Gets if the node is an AwaitExpression.
+     */
+    static isAwaitExpression: (node: Node<Node>) => node is AwaitExpression;
+    /**
+     * Gets if the node is a BinaryExpression.
+     */
+    static isBinaryExpression: (node: Node<Node>) => node is BinaryExpression<BinaryExpression>;
+    /**
+     * Gets if the node is a BindingElement.
+     */
+    static isBindingElement: (node: Node<Node>) => node is BindingElement;
+    /**
+     * Gets if the node is a Block.
+     */
+    static isBlock: (node: Node<Node>) => node is Block;
+    /**
+     * Gets if the node is a BooleanKeyword.
+     */
+    static isBooleanKeyword: (node: Node<Node>) => node is Expression<Expression>;
+    /**
+     * Gets if the node is a BreakStatement.
+     */
+    static isBreakStatement: (node: Node<Node>) => node is BreakStatement;
+    /**
+     * Gets if the node is a CallExpression.
+     */
+    static isCallExpression: (node: Node<Node>) => node is CallExpression<CallExpression>;
+    /**
+     * Gets if the node is a CaseBlock.
+     */
+    static isCaseBlock: (node: Node<Node>) => node is CaseBlock;
+    /**
+     * Gets if the node is a CaseClause.
+     */
+    static isCaseClause: (node: Node<Node>) => node is CaseClause;
+    /**
+     * Gets if the node is a CatchClause.
+     */
+    static isCatchClause: (node: Node<Node>) => node is CatchClause;
+    /**
+     * Gets if the node is a ClassDeclaration.
+     */
+    static isClassDeclaration: (node: Node<Node>) => node is ClassDeclaration;
+    /**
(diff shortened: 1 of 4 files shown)


dsherret:
Hey @lazarljubenovic, this is a good idea!

`isJSDocTag` is like that way because the nodes look like this in the compiler api:

```ts
interface JSDocTag extends Node {
    parent: JSDoc | JSDocTypeLiteral;
    tagName: Identifier;
    comment?: string;
}
interface JSDocUnknownTag extends JSDocTag {
    kind: SyntaxKind.JSDocTag;
}
/**
 * Note that `@extends` is a synonym of `@augments`.
 * Both tags are represented by this interface.
 */
interface JSDocAugmentsTag extends JSDocTag {
    kind: SyntaxKind.JSDocAugmentsTag;
    class: ExpressionWithTypeArguments & {
        expression: Identifier | PropertyAccessEntityNameExpression;
    };
}
interface JSDocAuthorTag extends JSDocTag {
    kind: SyntaxKind.JSDocAuthorTag;
}
//etc...
```

We'll need to get it to generate `isJSDocTag` back to the way it was, but other than that, this looks like a good change! 😄

Also, can you run `yarn overwrite-declaration-files` and commit in those changes? I wonder what the final declaration file looks like. It may be nice to pre-resolve the result of `ImplementedKindToNodeMappings[TKind]` if that's not done automatically (pre-resolving it helps speed things up and won't cause people to hit the conditional type depth limit as easily). If it doesn't pre-resolve it then we can just explicitly type everything with the code generation (ex. `static isAnyKeyword: ((node: compiler.Node) => node is compiler.Expression) = TypeGuards.is(SyntaxKind.AnyKeyword);` or whatever)

lazarljubenovic:
> pre-resolving it helps speed things up and won't cause people to hit the conditional type depth limit as easily

Ha! I love how the perspective of what are good practices to do at "build-time" and "run-time" completely change when you have a codegen tool at your fingertips. It makes you re-think everything you take for granted.

I changed the decision of "new-style* prop or old-style method" to also be based on the number of items in `syntaxKinds`, it seems like this is properly catching the `JSDocTag` which I previously let slid by by looking only at `isMixin`.

<sub>I don't mean "new/old" in any negative sense, of course; just trying to be clear. :smile:</sub>

Seems like it's being pre-resolved... not so nicely.

### Was

```ts
    /**
     * Gets if the node is an AnyKeyword.
     * @param node - Node to check.
     */
    static isAnyKeyword(node: Node): node is Expression;
    /**
     * Gets if the node is an ArrayLiteralExpression.
     * @param node - Node to check.
     */
    static isArrayLiteralExpression(node: Node): node is ArrayLiteralExpression;
```

### Is :roll_eyes: 

```
    /**
     * Gets if the node is an AnyKeyword.
     */
    static isAnyKeyword: (node: Node<Node>) => node is Expression<Expression>;
    /**
     * Gets if the node is an ArrayLiteralExpression.
     */
    static isArrayLiteralExpression: (node: Node<Node>) => node is ArrayLiteralExpression;
```

I had no idea that pre-resolving was even a thing, so I don't know why this is happening :smile:

dsherret:
Looks good to me. Thanks @lazarljubenovic! :)

lazarljubenovic:
Hmm, I thought I've seen that the tests were failing when I submitted the PR because of the `.d.ts` file. Apparently `Node<Node>` was killing it. Are you sure it was right to merge this yet?

dsherret:
@lazarljubenovic yeah, I missed that they were failing, but it was a quick fix by just explicitly typing it:

https://github.com/dsherret/ts-morph/commit/666febb4b6fce71729e1cc7f325e91ba84f958f8#diff-44f441ef5bb102628a5638f1af59a410

lazarljubenovic:
Oh, I see now that you've fixed it already!

dsherret:
Also, added a `readonly` keyword and expanded it to all node kinds (even though it's really only useful for the implemented ones). I didn't think about it until after the merge.

Anyway, thanks again for the PR! 🙂

Issue #728, opened 2019-10-05
TypeGuards isSemicolonToken, isInferKeyword negated will infer node as 'never'


cancerberoSgx:
This two typeguards, isSemicolonToken, isInferKeyword , when used with negation logic operator will infer the node has type 'never' which I think is not correct, as shown :

<img width="444" alt="Screen Shot 2019-10-05 at 3 05 46 PM" src="https://user-images.githubusercontent.com/112746/66258968-c72e2b80-e781-11e9-848e-d12a4f97f07e.png">

snippet : 

```ts
import { Project, TypeGuards } from 'ts-morph'

const project = new Project()
const file = project.createSourceFile('f1.ts', 'var a = 1')
var n = file.getDescendants()[2]
if (n && !TypeGuards.isSemicolonToken(n)) {
  var c = n.getKindName()
}
```

I think this could be a solution but I wanted to ask before PR:

```ts
    static readonly isSemicolonToken: (node: Node) => node is Node<ts.Token<SyntaxKind.SemicolonToken>>;
```

dsherret:
@cancerberoSgx oops! Yup, it should be that. Feel free to open a PR.

dsherret:
I fixed this along with #760. The fix was a bit more involved for this one than I thought. I really need to clean up the code generation code...

PR #731, opened 2019-10-06
Dev

cancerberoSgx:
https://github.com/dsherret/ts-morph/issues/728

### DIFF
--- a/src/compiler/ast/common/Node.ts
+++ b/src/compiler/ast/common/Node.ts
@@ -1299,7 +1299,7 @@ export class Node<NodeType extends ts.Node = ts.Node> {
     }
 
     /**
-     * Gets the line number at the start of the node.
+     * Gets the line number (1-indexed) of the start of the node in the *full* text of its source file. 
      * @param includeJsDocComments - Whether to include the JS doc comments or not.
      */
     getStartLineNumber(includeJsDocComments?: boolean) {
@@ -1307,7 +1307,8 @@ export class Node<NodeType extends ts.Node = ts.Node> {
     }
 
     /**
-     * Gets the line number of the end of the node.
+     * Gets the line number (1-indexed) of the end of the node in the *full* text of its source file. 
+     * Example: `node.getSourceFile().getFullText().split('\n')[node.getEndLineNumber() - 1]`
      */
     getEndLineNumber() {
         const sourceFileText = this._sourceFile.getFullText();
--- a/src/utils/TypeGuards.ts
+++ b/src/utils/TypeGuards.ts
@@ -854,7 +854,7 @@ export class TypeGuards {
     static isIndexedAccessTypeNode(node: compiler.Node): node is compiler.IndexedAccessTypeNode {
         return node.getKind() === SyntaxKind.IndexedAccessType;
     }
-    
+
     /**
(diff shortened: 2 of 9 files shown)


src/compiler/ast/common/Node.ts
```diff
@@ -1299,15 +1299,16 @@ export class Node<NodeType extends ts.Node = ts.Node> {
     }
 
     /**
-     * Gets the line number at the start of the node.
+     * Gets the line number (1-indexed) of the start of the node in the *full* text of its source file.
```
dsherret: Thanks! This is an improvement, but let's remove the emphasis (as nothing makes sense other than full text) and use "at" for both of these.

```suggestion
     * Gets the line number (1-indexed) at the start of the node in the full text of its source file. 
```

src/compiler/ast/common/Node.ts
```diff
@@ -1299,15 +1299,16 @@ export class Node<NodeType extends ts.Node = ts.Node> {
...
+     * Gets the line number (1-indexed) of the start of the node in the *full* text of its source file. 
      * @param includeJsDocComments - Whether to include the JS doc comments or not.
      */
     getStartLineNumber(includeJsDocComments?: boolean) {
         return StringUtils.getLineNumberAtPos(this._sourceFile.getFullText(), this.getStartLinePos(includeJsDocComments));
     }
 
     /**
-     * Gets the line number of the end of the node.
+     * Gets the line number (1-indexed) of the end of the node in the *full* text of its source file.
```
dsherret:
```suggestion
     * Gets the line number (1-indexed) at the end of the node in the full text of its source file. 
```

src/utils/TypeGuards.ts
```diff
@@ -857,7 +858,9 @@ export class TypeGuards {
     /**
      * Gets if the node is a InferKeyword.
      */
-    static readonly isInferKeyword: (node: compiler.Node) => node is compiler.Node = TypeGuards.is(SyntaxKind.InferKeyword);
+    static isInferKeyword(node: compiler.Node): node is compiler.Node<Token<SyntaxKind.InferKeyword>> { 
+        return node.getKind() === SyntaxKind.InferKeyword;
+    }
```
dsherret: This file is actually code generated so we can't make this change here. For example, this PR did the change to make everything in here use `TypeGuards.is`: https://github.com/dsherret/ts-morph/pull/703/files

Do you want to merge the jsdoc change above separately as that's a good change? Then we could come up for a fix for this separately. Let me know if you'd like to look at fixing the generated code or I can take a look at it soon.

cancerberoSgx: I saw the note in the header comment but thought only applied to the actual method is()

I will fix there then. 

do you agree with that type casting ?

dsherret: Oh yeah, I can see how that header comment might be confusing. Feel free to fix it!

Do you mean casting the result of `TypeGuards.is(SyntaxKind.InferKeyword)` to `compiler.Node<Token<SyntaxKind.InferKeyword>>`? That should be fine!

Thanks again! 😊

dsherret:
Hmmm... I think github automatically closed this (maybe because the folder structure has changed so drastically). Feel free to reopen later!

PR #743, opened 2019-10-18
Soft rename addExistingSourceFile(s)(IfExists)

### LINKED ISSUE #733: Soft rename `addExistingSourceFile` to `addSourceFileAtPath`
I always didn't like this method name. Just thought of this which I think is better.

This should be a soft rename for a long time to allow people ample time to switch.

Would also be nice to have code mods for these changes...

Todo: Consider either `addSourceFileAtPath/addSourceFileAtPathOrThrow` or `addSourceFileAtPath/addSourceFileAtPathIfExists`. Probably `IfExists` is better.

ChristianIvicevic:
This PR contains the soft rename and deprecation of `addExistingSourceFile`, `addExistingSourceFileIfExists` and `addExistingSourceFiles`.

Closes #733

### DIFF
--- a/packages/ts-morph/src/Project.ts
+++ b/packages/ts-morph/src/Project.ts
@@ -229,9 +229,31 @@ export class Project {
      * Adds source files based on file globs.
      * @param fileGlobs - File glob or globs to add files based on.
      * @returns The matched source files.
+     * @deprecated Use `addSourceFilesAtPaths`.
      */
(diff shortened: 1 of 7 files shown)


ChristianIvicevic:
The code coverage most likely decreased because I removed all references to the deprecated methods and thus they are no longer covered.

dsherret:
@ChristianIvicevic sweet! Thanks for doing this on the dev branch :)

PR #773, opened 2019-12-11
chore: fix package name

basarat:
Discovered the project a week ago :rose: :heart:


dsherret:
Thanks a lot, @basarat! I have very crude type checking going on for all the docs except for on that readme.md for the bootstrap project. I'll have to add that there soon...

Let me know if you can think of any improvements to that `@ts-morph/bootstrap` library.

PR #799, opened 2020-03-17
fix: rename anonymous class declaration

harttle:
The `addNameNode()` in src/compiler/ast/base/name/NameableNode.ts searches for `OpenParenToken` to add name node. For anonymous class declarations it should be `OpenBraceToken`, like:

```typescript
export default class {}
```


dsherret:
Thanks so much @harttle! Sorry for my delay on this.

dsherret:
I did a 7.0.1 release for this. Thanks again!

Issue #806, opened 2020-04-07
Apply file text changes performance problem


killagu:
If the source file is vary large, more than 200,000 lines,  the program will stuck at [getTextFromTextChanges](https://github.com/dsherret/ts-morph/blob/latest/packages/ts-morph/src/manipulation/formatting/getTextFromFormattingEdits.ts#L18).

```js
text = text.slice(0, span.getStart()) + edit.getNewText() + text.slice(span.getEnd());
```

If the text is very long, build a new string cost lot.

We can optimise it with a string array.

```js
const text = sourceFile.getFullText();
let start = 0;
let end = 0;
let edits = [];
for (const { edit } of reversedFormattingEdits) {
    const span = edit.getSpan();
    end = span.getStart();
    const beforeEdit = text.slice(start, end);
    start = end;
    edits.push(beforeEdit);
    edits.push(edit.getNewText());
}
edits.push(text.slice(end));
return edits.join('');
```

If it is ok, I can make a pr to solve the problem.

dsherret:
@killagu for sure! I do this same technique in other places.

killagu:
Would you mind to release a new version with the https://github.com/dsherret/ts-morph/pull/807? I have a project occurred this performance problem.

dsherret:
@killagu published in 7.0.2 🙂

PR #807, opened 2020-04-07
perf(manipulation): optimise getTextFromTextChanges performance

killagu:
Refs:
- https://github.com/dsherret/ts-morph/issues/806

### DIFF
--- a/packages/ts-morph/src/manipulation/formatting/getTextFromFormattingEdits.ts
+++ b/packages/ts-morph/src/manipulation/formatting/getTextFromFormattingEdits.ts
@@ -12,12 +12,20 @@ export function getTextFromTextChanges(sourceFile: SourceFile, textChanges: Read
         return difference > 0 ? 1 : -1;
     });
 
-    let text = sourceFile.getFullText();
+    const text = sourceFile.getFullText();
+    let start = 0;
+    let end = text.length;
+    const editResult: Array<string> = [];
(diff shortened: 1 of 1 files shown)


packages/ts-morph/src/manipulation/formatting/getTextFromFormattingEdits.ts
```diff
@@ -12,12 +12,20 @@ export function getTextFromTextChanges(sourceFile: SourceFile, textChanges: Read
...
-        text = text.slice(0, span.getStart()) + edit.getNewText() + text.slice(span.getEnd());
+        start = span.getEnd();
+        const afterResult = text.slice(start, end);
+        end = span.getStart();
+        editResult.push(afterResult);
+        editResult.push(edit.getNewText());
     }
-    return text;
+    editResult.push(text.slice(0, end));
+    return editResult.reverse().join('');
```
dsherret, line 28: If we're optimizing for performance, perhaps we don't have to do line 6 anymore and could remove this `.reverse()`? Thoughts?

It might be good to ensure the text changes are in order anyway.

Overall, this looks good!

killagu: Yes, reverse twice not make sense. I'll delete it.

killagu:
🎉🎉🎉

PR #841, opened 2020-07-27
fix: change newline charactor CRLF to LF

sisisin:
CRLF Character causes git diff always
I fixed this problem, thx.


dsherret:
@sisisin Ah! The missing \r\n -> \n. I was wondering where that was. Thanks a lot! 😀

Issue #847, opened 2020-07-29
"Clean" method


jeswr:
**Is your feature request related to a problem? Please describe.**

A generic method to automatically "clean" generated code (formatting imports etc.).

**Describe the solution you'd like**

The `cleanFile` function added as a method to source files and the `clean` function added as a method to projects.

```ts
import Crypto from 'crypto-js'

function hashProject (project: Project): string {
    return Crypto.MD5(project.getSourceFiles().map(node => node.print()).toString()).toString()
}

function cleanFile(file: SourceFile) {
    file.fixMissingImports()
    for (let printed; printed != (printed = file.print());) {
        file.fixUnusedIdentifiers()
        file.organizeImports()
    }
    file.formatText()
}

function clean(project: Project) {
    for (let hash; hash != (hash = hashProject(project));)
        project.getSourceFiles().forEach(cleanFile)
}
```
alternatively without the Crypto dependency
```ts
function cleanFile(file: SourceFile) {
    file.fixMissingImports()
    for (let val; val != (val = file.print());) {
        file.fixUnusedIdentifiers()
        file.organizeImports()
    }
    file.formatText()
}

function clean(project: Project) {
    for (let val; val != (val = project.getSourceFiles().map(node => node.print()).toString());)
        project.getSourceFiles().forEach(cleanFile)
}
```
**Describe alternatives you've considered**
I can just have this in the files for my project but it seems like it would be useful in this framework.

dsherret:
Thanks for the suggestion, but this is probably something that's too arbitrary for what it should contain for this to be in the library. I'd rather only expose functions found in the compiler API (so if there were a "clean" quick fix in the compiler API then it would be good to add it).

The compiler API's formatting could also use a huge improvement... I recommend using my [dprint](https://github.com/dprint/dprint) formatter over that 🙂

PR #848, opened 2020-07-29
Resolving issue https://github.com/dsherret/ts-morph/issues/847

jeswr:


dsherret:
Thanks for this, but closing due to reasons outlined here: https://github.com/dsherret/ts-morph/issues/847#issuecomment-671078248

PR #850, opened 2020-08-11
test(ts-morph): fix failured test case

sisisin:


dsherret:
I was confused about this, but now I figured it out. You correctly did the change on the `master` branch, which I forgot to update, then now this PR is targeting the `latest` branch, which has the change.

I will need to rethink my branching strategy because this is not working overall. I'm going to close this because it was already fixed here: https://github.com/dsherret/ts-morph/commit/3f8c40f53587ced9f3c619fb5eec18789a1be112

PR #853, opened 2020-08-13
Fix nested sibling transformations

### LINKED ISSUE #852: transform doesn't handle multiple nested transformed siblings
**Describe the bug**

Version: 7.3.0

https://github.com/dsherret/ts-morph/blob/51cbbea3da79b54e50afd9bec0cfb43f8737d073/packages/ts-morph/src/compiler/ast/common/Node.ts#L1478-L1480

should repeatedly `pop` nested transformations (instead of just once)

**To Reproduce**

```ts
import { Project, ts } from "ts-morph";

console.log(ts.version); // 3.9.7

const sourceFile = project.createSourceFile("test.ts", `foo(bar(), baz())`);

sourceFile.transform(traversal => {
    const node = traversal.visitChildren();
    if (ts.isCallExpression(node)) {
        // Replace every `foo(...)` with `hello`
        return ts.createIdentifier("hello");
    }
    return node;
});

console.log(sourceFile.getFullText());
```

**Expected behavior**

```ts
hello
```

**Actual behavior**

```ts
foo(hellofoo(bar()hello
```

vsiao:
Fixes #852 
Also added a test; confirmed that it fails prior to changes:

image

### DIFF
--- a/packages/ts-morph/src/compiler/ast/common/Node.ts
+++ b/packages/ts-morph/src/compiler/ast/common/Node.ts
@@ -1438,7 +1438,8 @@ export class Node<NodeType extends ts.Node = ts.Node> {
             newLine: this._context.manipulationSettings.getNewLineKind(),
             removeComments: false,
         });
-        const transformations: { start: number; end: number; compilerNode: ts.Node; }[] = [];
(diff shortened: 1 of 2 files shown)


dsherret:
Looks great. Thanks a lot @vsiao!

PR #875, opened 2020-10-09
fix: fixed adding const modifier to exported enum

IKatsuba:


dsherret:
@IKatsuba thanks a lot for this fix! Looks great and this is now published in 8.1.2

IKatsuba:
@dsherret, you are welcome! I needed to fix it!

PR #883, opened 2020-11-19
preserve files outside rootDirs. fix #878

### LINKED ISSUE #878: definition files not under rootDirs are excluded unexpectedly
**Describe the bug**

Version: 0.6.0

**To Reproduce**

1. `git clone https://github.com/Zzzen/morph-demo`
2. `node morph.js`

main files:

```ts
// tsconfig.json
{
  "compilerOptions": {
    /* Basic Options */
    "outDir": "build",
    "rootDirs": ["src"],                        /* List of root folders whose combined content represents the structure of the project at runtime. */
  },
  "include": ["types.d.ts", "src/*.ts"]
}


// types.d.ts
declare const SOME_GLOBAL_VAR: string;

// src/index.ts
const foo = SOME_GLOBAL_VAR;

// morph.js
const { createProject, ts } = require("@ts-morph/bootstrap");
main();
async function main() {
    const project = await createProject({ tsConfigFilePath: "tsconfig.json" });
    const diagnostics = ts.getPreEmitDiagnostics(project.createProgram());
    console.log(diagnostics);
}

```

**Expected behavior**

like how tsc compiles, there should be no diagnostics.

`typed.d.ts` is specified in `include` field of tsconfig. But it's discarded by [TsConfigResolver](https://github.com/dsherret/ts-morph/blob/b8bbeb5410d0e7b7205fb4923d0e2f346147f368/packages/common/src/tsconfig/TsConfigResolver.ts#L47) because it is not a child of any rootDirs.

Zzzen:
https://www.typescriptlang.org/tsconfig#rootDir

>Importantly, rootDir does not affect which files become part of the compilation. It has no interaction with the include, exclude, or files tsconfig.json settings.

Neither rootDir nor rootDirs affect file inclusions. `include` is `[]` if `files` is specified, otherwise `["**/*"]`.

### DIFF
--- a/packages/common/src/tsconfig/TsConfigResolver.ts
+++ b/packages/common/src/tsconfig/TsConfigResolver.ts
@@ -44,8 +44,8 @@ export class TsConfigResolver {
         for (let fileName of configFileContent.fileNames) {
             const filePath = fileSystem.getStandardizedAbsolutePath(fileName);
             const parentDirPath = FileUtils.getDirPath(filePath);
+            files.add(filePath);
             if (dirInProject(parentDirPath) && fileSystem.fileExistsSync(filePath)) {
-                files.add(filePath);
                 directories.add(parentDirPath);
             }
         }
--- a/packages/ts-morph/src/tests/projectTests.ts
+++ b/packages/ts-morph/src/tests/projectTests.ts
@@ -27,7 +27,7 @@ describe(nameof(Project), () => {
 
         it("should add the files from tsconfig.json by default with the target in the tsconfig.json", () => {
             const fileSystem = new InMemoryFileSystemHost({ skipLoadingLibFiles: true });
-            fileSystem.writeFileSync("tsconfig.json", `{ "compilerOptions": { "rootDir": "test", "target": "ES5" } }`);
+            fileSystem.writeFileSync("tsconfig.json", `{ "compilerOptions": { "rootDir": "test", "target": "ES5" }, "include": ["test"] }`);
             fileSystem.writeFileSync("/otherFile.ts", "");
             fileSystem.writeFileSync("/test/file.ts", "");
             fileSystem.writeFileSync("/test/test2/file2.ts", "");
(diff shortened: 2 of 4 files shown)


packages/common/src/tsconfig/TsConfigResolver.ts
```diff
@@ -44,8 +44,8 @@ export class TsConfigResolver {
         for (let fileName of configFileContent.fileNames) {
             const filePath = fileSystem.getStandardizedAbsolutePath(fileName);
             const parentDirPath = FileUtils.getDirPath(filePath);
+            files.add(filePath);
             if (dirInProject(parentDirPath) && fileSystem.fileExistsSync(filePath)) {
-                files.add(filePath);
                 directories.add(parentDirPath);
```
Zzzen, line 49: What are `directories` used for? No usage is found in this project.

packages/ts-morph/src/tests/projectTests.ts
```diff
@@ -43,7 +43,7 @@ describe(nameof(Project), () => {
             fileSystem.writeFileSync("/test/file.ts", "");
             fileSystem.writeFileSync("/test/test2/file2.ts", "");
             const project = new Project({ tsConfigFilePath: "tsconfig.json", compilerOptions: { rootDir: "/test/test2" }, fileSystem });
-            expect(project.getSourceFiles().map(s => s.getFilePath()).sort()).to.deep.equal(["/test/test2/file2.ts"].sort());
+            expect(project.getSourceFiles().map(s => s.getFilePath()).sort()).to.deep.equal(["/otherFile.ts", "/test/file.ts", "/test/test2/file2.ts"].sort());
```
Zzzen, line 46: `include` is `["**/*"]` by default, so `/otherFile.ts` and `/test/file.ts` should be included too. And tsc should fail successfully.

> error TS6059: File '/otherFile.ts' is not under 'rootDir' '/test/test2'. 'rootDir' is expected to contain all source files.

packages/common/src/tsconfig/TsConfigResolver.ts
dsherret: It's used in the ts-morph project (not `@ts-morph/bootstrap`).

packages/ts-morph/src/tests/projectTests.ts
dsherret: Learn something new every day 😄

packages/common/src/tsconfig/TsConfigResolver.ts
dsherret: I think maybe this should always add the directory as well, but I will look into this after merge.

dsherret:
Thanks @Zzzen!

dsherret:
I looked into this a bit more and did an additional commit to fix things up (https://github.com/dsherret/ts-morph/commit/8c7e8ba2e7d6a4a1ae36d0642c88f770ac1ee905). Let me know if there's anything wrong in there (going to do a release later today with this).

By the way... new tsconfig.json documentation is sooo much better for understanding this... hadn't looked at it in a while. Orta did a good job!

Zzzen:
Thank you for the quick response. The tests are cleaner now.

Issue #881, opened 2020-11-13
Implement getProjectVersion on LanguageServiceHost


geremih:
Thanks for creating this awesome project!

**Is your feature request related to a problem? Please describe.**
I have a script that extracts some type information which was taking longer than I would have expected. On profiling the script the bottleneck appears to be calls to `synchronizeHostData` which are made by most calls to the language-service eg. [getDefinitionAtPosition](https://github.com/microsoft/TypeScript/blob/566faa7a01f3d483f032fda4777dae186eccbb19/src/services/services.ts#L1634).

`synchronizeHostData`  short-circuits if the host supports a `getProjectVersion` and the project version hasn't changed between calls. On implementating a dummy `getProjectVersion` the run time of my script went done from 150s to 10s.

Related: [original implementation in typescript](https://github.com/microsoft/TypeScript/pull/3131), [ts-loader PR that adds support for getProjectVersion](https://github.com/TypeStrong/ts-loader/pull/100)

**Describe the solution you'd like**
I'm up for contributing this myself if the plan sounds good to you. The general steps I had in mind are:
1. Add a version field on `TransactionalFileSystem` that is incremented on mutating operations (writeFile, deleteFile) etc.
2. (Optionally) Add a field to `ProjectOptions` to enable this behaviour to support the project being mutating outside ts-morph calls.

dsherret:
@geremih Sorry for my delay—I've been super busy lately. This sounds like a good thing to do for sure!

I don't think this should be added to `TransactionalFileSystem` though because changes to a source file might not have hit the transactional file system yet.

I believe this would be super simple. I think the solution might be to just rename [`LanguageService#_resetProgram`](https://github.com/dsherret/ts-morph/blob/4ed90e0a4d69c3eeab703a7caad5514445d67820/packages/ts-morph/src/compiler/tools/LanguageService.ts#L66) to `LanguageService#_reset` then make it increment a value in there that's then returned in the `getProjectVersion()` function on the language service host. See how language service host is created [here](https://github.com/dsherret/ts-morph/blob/4ed90e0a4d69c3eeab703a7caad5514445d67820/packages/ts-morph/src/compiler/tools/LanguageService.ts#L41) and [here](https://github.com/dsherret/ts-morph/blob/4ed90e0a4d69c3eeab703a7caad5514445d67820/packages/common/src/compiler/createHosts.ts#L27). Let me know if I'm mistaken about anything.

If you'd like to implement this, then please go for it! 🙂

[...]

PR #884, opened 2020-11-19
Implement getProjectVersion on LanguageServiceHost

geremih:
Fixes https://github.com/dsherret/ts-morph/issues/881

### DIFF
--- a/packages/ts-morph/src/ProjectContext.ts
+++ b/packages/ts-morph/src/ProjectContext.ts
@@ -29,7 +29,6 @@ export class ProjectContext {
     private readonly _compilerOptions: CompilerOptionsContainer;
     private readonly _customTypeChecker: TypeChecker | undefined;
     private readonly _project: Project | undefined;
-    private _projectVersion: number  = 0;
 
     get project(): Project {
         if (this._project == null)
@@ -68,8 +67,6 @@ export class ProjectContext {
             this._customTypeChecker = new TypeChecker(this);
             this._customTypeChecker._reset(() => opts.typeChecker!);
         }
-        this.compilerFactory.onSourceFileAdded(() => this._projectVersion += 1);
-        this.compilerFactory.onSourceFileRemoved(() => this._projectVersion += 1);
     }
 
     /** Gets the compiler options. */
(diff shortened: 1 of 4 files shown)


packages/ts-morph/src/ProjectContext.ts
```diff
@@ -29,6 +29,7 @@ export class ProjectContext {
     private readonly _compilerOptions: CompilerOptionsContainer;
     private readonly _customTypeChecker: TypeChecker | undefined;
     private readonly _project: Project | undefined;
+    private _projectVersion: number  = 0;
```
dsherret, line 32: Perhaps since this only relates to the language service, let's move this over to `LanguageService`.

packages/ts-morph/src/ProjectContext.ts
```diff
@@ -67,6 +68,8 @@ export class ProjectContext {
             this._customTypeChecker = new TypeChecker(this);
             this._customTypeChecker._reset(() => opts.typeChecker!);
         }
+        this.compilerFactory.onSourceFileAdded(() => this._projectVersion += 1);
+        this.compilerFactory.onSourceFileRemoved(() => this._projectVersion += 1);
```
dsherret, line 72: I'm looking at the TS source code and wow, you're right that this will really speed things up! Question though, does this not need to occur when a source file is changed? I might be missing something though.

dsherret (changes requested):
I'm really excited for this one and thanks a lot for the PR! I'm wondering though if this needs to also occur when a source file is changed? It seems to be that way, but maybe I'm misreading the code. If so, we might be able to just rename `LanguageService#_resetProgram` to `LanguageService#_reset` and then put the increment in there.

packages/ts-morph/src/ProjectContext.ts
geremih: Ah yeah, I missed the call to reset from the ProjectContext.

geremih:
@dsherret Thanks for the review. Have updated the PR

dsherret (approved):

dsherret:
Awesome. Thanks @geremih! I'm going to do a new release soon.

PR #893, opened 2020-12-05
Add getPosAtLineNumber

### LINKED ISSUE #891: Get node at (line, col)
**Is your feature request related to a problem? Please describe.**
I get line & column coordinates from an external source, I want to get the node(s) at this coordinate.

**Describe the solution you'd like**
The obvious solution would be `sourceFile.getChildAtPos`, but this takes a 1D coordinate. I can convert from "pos" to line+col using `sourceFile.getLineAndColumnAtPos` (#801) but I can't find anything for the reverse.

**Describe alternatives you've considered**
I bodged together this:
```ts
source.forEachDescendant(node => {
    const { line, column } = source.getLineAndColumnAtPos(node.getPos());
    if (line === searchLine && column === searchColumn) {
        // Node found!
    }
})
```
But that doesn't seem like an efficient solution.

HoldYourWaffle:
I have never encountered as many off-by-one-ish errors as I have today...
However, as far as I can tell this function is now working properly 😄

I also changed a comment somewhere to be more clear about *what* is 1-indexed.

Closes #891.

### DIFF
--- a/packages/common/lib/ts-morph-common.d.ts
+++ b/packages/common/lib/ts-morph-common.d.ts
@@ -966,6 +966,7 @@ export declare class StringUtils {
     static endsWithNewLine(str: string | undefined): boolean;
     static insertAtLastNonWhitespace(str: string, insertText: string): string;
     static getLineNumberAtPos(str: string, pos: number): number;
+    static getPosAtLineNumber(str: string, line: number): number;
     static getLengthFromLineStartAtPos(str: string, pos: number): number;
     static getLineStartFromPos(str: string, pos: number): number;
     static getLineEndFromPos(str: string, pos: number): number;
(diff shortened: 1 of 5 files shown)


Gerrit0:
Did you know about the `getPositionOfLineAndCharacter` function provided by TypeScript? It will perform significantly better over repeated calls since it caches the line positions, and avoids copying strings.

dsherret:
Thanks @Gerrit0 for that suggestion! I didn't know about that.

@HoldYourWaffle what do you think about updating this PR to use that API? It might be nice if it caches stuff for performance and reduces allocations. I think it might be 0-indexed, but ts-morph is 1-indexed.

HoldYourWaffle:
Sounds good! I'll hopefully get to it later today.

PR #905, opened 2020-12-15
docs(ts-morph): clarify behaviour of getTargetType

### LINKED ISSUE #904: make getTargetType return undefined if would return self
**Describe the bug**

Version: 9.1.0

**To Reproduce**

```ts
import dedent from 'dedent'
import * as tsm from 'ts-morph'
import { dumpType } from './utils'

const tsMorphProject = new tsm.Project({
  skipAddingFilesFromTsConfig: true,
  useInMemoryFileSystem: true,
})

tsMorphProject.createSourceFile(
  '/repro-example.ts',
  dedent`
    export interface Foo<T> {}
  `
)

tsMorphProject.addSourceFileAtPath('/repro-example.ts')

tsMorphProject.getSourceFiles().map((sf) => {
  sf.getExportedDeclarations().forEach((ex) => {
    const t = ex[0]!.getType()!
    dumpType(t)
  })
})
```

My log:

```
t.getText()                                          = import("/repro-example").Foo<T>
t.getSymbol()?.getName()                             = Foo
t.compilerType.getFlags()                            = 524288
(t.compilerType as tsm.ts.ObjectType)?.objectFlags   = 6
t.getAliasSymbol()?.getName()                        = undefined
t.getApparentType().getText()                        = import("/repro-example").Foo<T>
t.getSymbol()?.getDeclarations()?.[0]?.getKindName() = InterfaceDeclaration
t.getSymbol()?.getDeclarations()?.[0]?.getText()     = export interface Foo<T> {}

t.isAnonymous()                                      = false
t.isAny()                                            = false
t.isInterface()                                      = true
t.isObject()                                         = true
t.isArray()                                          = false
t.isBoolean()                                        = false
t.isBooleanLiteral()                                 = false
t.isClass()                                          = false
t.isClassOrInterface()                               = true
t.isEnum()                                           = false

t.getAliasTypeArguments().length}                    = 0
Boolean(t.getTargetType())                           = true
t.getTargetType() === t                              = true

t.getText(undefined, tsm.ts.TypeFormatFlags.InTypeAlias)                           = Foo<T>
t.getText(undefined, tsm.ts.TypeFormatFlags.UseAliasDefinedOutsideCurrentScope)    = Foo<T>
t.getText(undefined, tsm.ts.TypeFormatFlags.UseTypeOfFunction)                     = Foo<T>
t.getText(undefined, tsm.ts.TypeFormatFlags.UseFullyQualifiedType)                 = import("/repro-example").Foo<T>
t.getText(undefined, tsm.ts.TypeFormatFlags.NoTruncation)                          = Foo<T>
t.getText(undefined, tsm.ts.TypeFormatFlags.UseStructuralFallback)                 = Foo<T>

Target Type?
------------
N/A
```

Notice that I mark `Foo` has having no target type here (N/A) because the target type refers to self via reference equality.

 

**Expected behavior**

This looks like a bug to me. It seems that calling `getTargetType` in this case should return `undefined` (or maybe slightly better `null`). It seems to me that `getTargetType` should recursively return values until the terminal non-instantiated type-with-generics is reached.

jasonkuhrt:
closes #904

Instigated by https://github.com/dsherret/ts-morph/issues/904#issuecomment-745349988

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,4 @@
 {
-  "typescript.tsdk": "./node_modules/typescript/lib"
+  "typescript.tsdk": "./node_modules/typescript/lib",
+  "prettier.enable": false
 }
--- a/packages/ts-morph/src/compiler/types/Type.ts
+++ b/packages/ts-morph/src/compiler/types/Type.ts
@@ -243,9 +243,9 @@ export class Type<TType extends ts.Type = ts.Type> {
      *
      * For example:
      *
-     * - given `Promise<string>` this will return `Promise<T>`.
-     * - given `Promise<T>` this will return the same `Promise<T>`.
-     * - given `string` this will return `undefined`.
+     * - Given `Promise<string>` this will return `Promise<T>`.
+     * - Given `Promise<T>` this will return the same `Promise<T>`.
+     * - Given `string` this will return `undefined`.
      */
     getTargetType(): Type<ts.GenericType> | undefined {
         const targetType = (this.compilerType as any as ts.TypeReference).target || undefined;
--- a/packages/ts-morph/src/compiler/types/Type.ts
+++ b/packages/ts-morph/src/compiler/types/Type.ts
@@ -233,19 +233,19 @@ export class Type<TType extends ts.Type = ts.Type> {
     }
 
     /**
-     * Gets the reference type if this type is an instantiated type.
+     * Gets the "target type".
      *
      * Specifically:
      *
-     * - If this type is an instantiated type then returns the reference type.
-     * - If this type is a reference type then returns itself (same reference).
+     * - If the type is a type reference then returns the generic type.
+     * - If the type is already a generic type then returns itself (_same_ reference).
      * - Otherwise returns `undefined`.
      *
      * For example:
      *
-     * - Given `Promise<string>` this will return `Promise<T>`.
-     * - Given `Promise<T>` this will return the same `Promise<T>`.
-     * - Given `string` this will return `undefined`.
+     * - Given type reference `Promise<string>` returns `Promise<T>`.
+     * - Given generic type `Promise<T>` returns the same `Promise<T>`.
+     * - Given `string` returns `undefined`.
      */
     getTargetType(): Type<ts.GenericType> | undefined {
         const targetType = (this.compilerType as any as ts.TypeReference).target || undefined;
--- a/packages/ts-morph/src/compiler/types/Type.ts
+++ b/packages/ts-morph/src/compiler/types/Type.ts
@@ -233,9 +233,19 @@ export class Type<TType extends ts.Type = ts.Type> {
     }
 
     /**
-     * Gets the target type of a type reference if it exists.
+     * Gets the reference type if this type is an instantiated type.
      *
-     * For example, given `Promise<string>` this will return `Promise<T>`.
+     * Specifically:
+     *
+     * - If this type is an instantiated type then returns the reference type.
+     * - If this type is a reference type then returns itself (same reference).
+     * - Otherwise returns `undefined`.
+     *
+     * For example:
+     *
+     * - given `Promise<string>` this will return `Promise<T>`.
+     * - given `Promise<T>` this will return the same `Promise<T>`.
+     * - given `string` this will return `undefined`.
      */
     getTargetType(): Type<ts.GenericType> | undefined {
         const targetType = (this.compilerType as any as ts.TypeReference).target || undefined;
--- a/packages/ts-morph/src/compiler/types/Type.ts
+++ b/packages/ts-morph/src/compiler/types/Type.ts
@@ -233,13 +233,8 @@ export class Type<TType extends ts.Type = ts.Type> {
     }
 
     /**
-     * Gets the "target type".
-     *
-     * Specifically:
-     *
-     * - If the type is a type reference then returns the generic type.
-     * - If the type is already a generic type then returns itself (_same_ reference).
-     * - Otherwise returns `undefined`.
+     * Returns the generic type when the type is a type reference, returns itself when it's
+     * already a generic type, or otherwise returns undefined.
      *
      * For example:
      *
@@ -253,10 +248,16 @@ export class Type<TType extends ts.Type = ts.Type> {
     }
 
     /**
-     * Gets the target type of a type reference or throws if it doesn't exist.
+     * Returns the generic type when the type is a type reference, returns itself when it's
+     * already a generic type, or otherwise throws an error.
      *
-     * For example, given `Promise<string>` this will return `Promise<T>`.
+     * For example:
+     *
+     * - Given type reference `Promise<string>` returns `Promise<T>`.
+     * - Given generic type `Promise<T>` returns the same `Promise<T>`.
+     * - Given `string` throws an error.
      */
+
     getTargetTypeOrThrow(): Type<ts.GenericType> {
         return errors.throwIfNullOrUndefined(this.getTargetType(), "Expected to find the target type.");
     }


packages/ts-morph/src/compiler/types/Type.ts
```diff
@@ -233,9 +233,19 @@ export class Type<TType extends ts.Type = ts.Type> {
...
+     *
+     * - If this type is an instantiated type then returns the reference type.
+     * - If this type is a reference type then returns itself (same reference).
+     * - Otherwise returns `undefined`.
+     *
+     * For example:
+     *
+     * - given `Promise<string>` this will return `Promise<T>`.
+     * - given `Promise<T>` this will return the same `Promise<T>`.
+     * - given `string` this will return `undefined`.
```
jasonkuhrt, line 248: Before Writing the JSdoc for the throws version I want to align on this one. The other will mostly be a copy except `return undefined` being replaced with `throws` ish.

packages/ts-morph/src/compiler/types/Type.ts
```diff
@@ -233,9 +233,19 @@ export class Type<TType extends ts.Type = ts.Type> {
...
+     *
+     * - If this type is an instantiated type then returns the reference type.
+     * - If this type is a reference type then returns itself (same reference).
+     * - Otherwise returns `undefined`.
+     *
+     * For example:
+     *
+     * - Given `Promise<string>` this will return `Promise<T>`.
+     * - Given `Promise<T>` this will return the same `Promise<T>`.
+     * - Given `string` this will return `undefined`.
```
jasonkuhrt, line 248: Before Writing the JSdoc for the throws version I want to align on this one. The other will mostly be a copy except `return undefined` being replaced with `throws` ish.

jasonkuhrt: > glossary that explains these terms/concepts in detail.

Maybe we should start one 😅

packages/ts-morph/src/compiler/types/Type.ts
```diff
@@ -233,9 +233,19 @@ export class Type<TType extends ts.Type = ts.Type> {
     }
 
     /**
-     * Gets the target type of a type reference if it exists.
+     * Gets the reference type if this type is an instantiated type.
```
dsherret, line 236: Is a "reference type" the same thing as "type reference"? In this case it's getting the target type from a type reference. Here's some docs in the compiler API:

https://github.com/microsoft/TypeScript/blob/4a9e2be386d558cf66d4bd90591c49498419bcf0/lib/typescript.d.ts#L2534

I guess we need different language for what "target type" means. Seems like we could maybe say "Gets the generic type of a type reference."?

packages/ts-morph/src/compiler/types/Type.ts
```diff
@@ -233,9 +233,19 @@ export class Type<TType extends ts.Type = ts.Type> {
...
      *
-     * For example, given `Promise<string>` this will return `Promise<T>`.
+     * Specifically:
+     *
+     * - If this type is an instantiated type then returns the reference type.
+     * - If this type is a reference type then returns itself (same reference).
+     * - Otherwise returns `undefined`.
```
dsherret, line 242: Maybe: "Otherwise returns `undefined` when not a type reference."

packages/ts-morph/src/compiler/types/Type.ts
dsherret: I just looked into what the terms may be:

https://github.com/microsoft/TypeScript/blob/4a9e2be386d558cf66d4bd90591c49498419bcf0/lib/typescript.d.ts#L2534

Maybe let's use "generic type" to refer to the target type and "type reference" to refer to the instantiated type so it aligns with the compiler api. We can use `Promise<T>` as an example of a generic type and `Promise<string>` as an example of a type reference. Sound good?

dsherret: @jasonkuhrt yeah, a lot of stuff in ts-morph is undocumented because I haven't understood a lot of stuff if the compiler api haha.

packages/ts-morph/src/compiler/types/Type.ts
jasonkuhrt: Technically its "when not a type reference OR a generic type". I feel it should be either totally explicit or implicit, as sayig in-between might confuse the message?

.vscode/settings.json
```diff
@@ -1,3 +1,4 @@
 {
-  "typescript.tsdk": "./node_modules/typescript/lib"
+  "typescript.tsdk": "./node_modules/typescript/lib",
+  "prettier.enable": false
```
jasonkuhrt, line 3: Whoops, will remove this.

packages/ts-morph/src/compiler/types/Type.ts
```diff
@@ -233,9 +233,19 @@ export class Type<TType extends ts.Type = ts.Type> {
...
+     *
+     * For example:
+     *
+     * - Given `Promise<string>` this will return `Promise<T>`.
+     * - Given `Promise<T>` this will return the same `Promise<T>`.
+     * - Given `string` this will return `undefined`.
      */
```
jasonkuhrt, line 249: I think a runnable `@example` block could be great. Something like this for example:

```ts
const exps = new tsm.Project({ useInMemoryFileSystem: true })
  .createSourceFile(
    'example.ts',
    `
      export type Generic<T> = {}
      export const foo: Generic<string> = {}
      export const bar: string = ''
    `
  )
  .getExportedDeclarations()

exps.get('Generic')[0].getType().getTargetType() // ?
exps.get('foo')[0].getType().getTargetType() === exps.get('Generic')[0].getType()
exps.get('bar')[0].getType().getTargetType() === undefined
```

Note for some reason the exported `Generic` target type is not returning self right now, but rather `undefined` which is not what I expected. will look into it later.

Do you think this or something like it would be a good idea and helpful for users?

packages/ts-morph/src/compiler/types/Type.ts
dsherret: Yes, true!

packages/ts-morph/src/compiler/types/Type.ts
```diff
@@ -233,9 +233,19 @@ export class Type<TType extends ts.Type = ts.Type> {
     }
 
     /**
-     * Gets the target type of a type reference if it exists.
+     * Gets the "target type".
```
dsherret, line 236: Sorry, this is having a bit of churn, but maybe we can put most of this information right in one or two sentences? Something like "Gets the generic type when the type is a type reference, returns itself when it's already the generic type, or otherwise returns undefined."?

I think we can get rid of the "Specifically" part if we just do that. I always find reading lists to understand functionality a bit harder to understand, but maybe it's just me.

packages/ts-morph/src/compiler/types/Type.ts
dsherret: Thanks for the suggestion, but I'd rather not at this moment. Updating examples for every scenario becomes too much to maintain. I think they will get the idea from the `Promise<string>` -> `Promise<T>` example in the "For example" section above.

.vscode/settings.json
dsherret: You could keep it if you'd like. I think there's no harm having this.

packages/ts-morph/src/compiler/types/Type.ts
jasonkuhrt: Yeah that's perfect!

packages/ts-morph/src/compiler/types/Type.ts
```diff
@@ -233,20 +233,31 @@ export class Type<TType extends ts.Type = ts.Type> {
     }
 
     /**
-     * Gets the target type of a type reference if it exists.
+     * Returns the generic type when the type is a type reference, returns itself when it's
+     * already a generic type, or otherwise returns undefined.
```
jasonkuhrt, line 237:
```suggestion
     * Gets the generic type when the type is a type reference, itself when it's
     * already a generic type, or otherwise returns undefined.
```

Better?

dsherret:
@jasonkuhrt looks great and thanks for adding this!

jasonkuhrt:
@dsherret thanks for discussing it!

PR #909, opened 2020-12-19
feat: add isKnownTypesPackageName option

chenjigeng:
This PR is to fix this [problem](https://github.com/microsoft/TypeScript/issues/41962).
When the host. isKnownTypesPackageName is undefined, languageService.getCodeFixesAtPosition will throw a error

### DIFF
--- a/packages/common/src/compiler/createHosts.ts
+++ b/packages/common/src/compiler/createHosts.ts
@@ -21,14 +21,17 @@ export interface CreateHostsOptions {
     /** Provides the current project version to be used to tell if source files have
      * changed. Provide this for a performance improvement. */
     getProjectVersion?: () => string;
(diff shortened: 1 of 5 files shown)


packages/common/src/compiler/createHosts.ts
```diff
@@ -65,7 +68,8 @@ export function createHosts(options: CreateHostsOptions) {
                 );
             }
         },
-        useCaseSensitiveFileNames: () => true,
+        isKnownTypesPackageName: isKnownTypesPackageName || returnTrue,
```
dsherret, line 71: I seem to not be able to push commits to this PR, so I'm going to merge this as-is, but afterwards I'm going to remove this `|| returnTrue` since I'd rather have undefined be the default as is done in the compiler api.

dsherret:
Thanks @chenjigeng!

PR #920, opened 2021-01-08
Add missing semicolons in lib.es2019.array.d.ts

FranklinWhale:
Currently 2 semicolons are missing in lib.es2019.array.d.ts so that when the file is used with the `InMemoryFileSystemHost` the compiler throws errors.

This PR should fix the errors.


dsherret:
Oh, interesting! This is a bug in https://github.com/dsherret/dts-minify though and that file is autogenerated by https://github.com/dsherret/ts-morph/blob/latest/packages/common/scripts/createLibFile.ts

Looks like we'll need to fix this in dts-minify, but I'm not sure how (looking into a fix now).

PR #1150, opened 2021-04-26
Expose `Project::options`

wojpawlik:
I need to parse `tsconfig.json` to access a [`"deno2node"` top-level key](https://github.com/wojpawlik/deno2node/issues/4#issuecomment-826314717). This tiny change would save me from having to pass `tsConfigFilePath` alongside `project`.

### DIFF
--- a/packages/ts-morph/src/Project.ts
+++ b/packages/ts-morph/src/Project.ts
@@ -65,7 +65,7 @@ export class Project {
      * Initializes a new instance.
      * @param options - Optional options.
      */
-    constructor(options: ProjectOptions = {}) {
+    constructor(readonly options: ProjectOptions = {}) {
         verifyOptions();
 
         // setup file system


dsherret:
@wojpawlik could you explain a bit more why this is necessary? (Maybe some example code)

What advantage does this have over the following?

```ts
const options: ProjectOptions = {
    // options go here...
};
const project = new Project(options);

// use options here...
```

wojpawlik:
It makes it easier to split the code into functions:

```ts
function createPonyfiller(project: Project, tsConfigFilePath: string) {
```

The function needs `tsConfigFilePath` to
1. locate (and parse) `tsconfig.json`,
2. resolve paths relatively to it.

dsherret:
I don't think it's up to ts-morph to hold onto this state because it's data already provided by the caller.

I'd recomend either doing something like this:

```ts
const tsConfigFilePath = "...";
const project = new Project({ tsConfigFilePath });

createPonyfiller(project, tsConfigFilePath);
```

Or:

```ts
const tsConfigFilePath = "...";
const project = new Project({ tsConfigFilePath });
const context = {
    project,
    tsConfigFilePath,
};

createPonyfiller(context);
```

Or:

```ts
const tsConfigFilePath = "...";
const project = new Project({ tsConfigFilePath });
const tsConfig = parseTsConfig(tsConfigFilePath);

const context = {
    project,
    tsConfig,
};

createPonyfiller(context);
```

PR #1158, opened 2021-05-12
fix - Handle comma separated definitions with DefinitionInfo#getDeclarationNode

IgorMinar:
The current implementation has an off-by-one bug that causes getDeclarationNode()
to return undefined instead of the defition when the definition is immediately
preceeded by a "," (comma token).

For example, "myFunction" in: `const someConst=23,myFunction=function () {};`

This change adjusts the lower bound of the AST travesal to not include the end
position, which in the case of comma separated variable declarations with no
whitespace results in the recursion incorrectly identifying the comma token as
the declaration node.

### DIFF
--- a/packages/ts-morph/src/compiler/tools/results/DefinitionInfo.ts
+++ b/packages/ts-morph/src/compiler/tools/results/DefinitionInfo.ts
@@ -63,7 +63,7 @@ export class DefinitionInfo<TCompilerObject extends ts.DefinitionInfo = ts.Defin
                 return node;
 
             for (const child of node._getChildrenIterator()) {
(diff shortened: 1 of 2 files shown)


dsherret:
Thanks @IgorMinar!

IgorMinar:
Thank you for ts-morph, it's been a pleasure to use instead of the TS APIs.

dsherret:
@IgorMinar thanks, I'm glad you like it! I haven't been working on it enough the past couple of years because I've been focusing on something else, but I think in the next month or so I'm going to be spending more time on ts-morph.

By the way, this is released now in 10.1.0.

IgorMinar:
Thank you for the release!

PR #1186, opened 2021-08-07
Update Doc: Navigation/get-source-files [add current working directory note]

MohamedLamineAllal:
Add a mention that the matching is done relatively to the current directory


dsherret:
Thanks @MohamedLamineAllal!

PR #1189, opened 2021-08-14
Avoid imporing code using unstable APIs

wojpawlik:
Works around https://github.com/denoland/deno_std/issues/1015.

I updated `std` while at it.

I don't know how to test this.

### DIFF
--- a/packages/common/src/runtimes/DenoRuntime.ts
+++ b/packages/common/src/runtimes/DenoRuntime.ts
@@ -1,7 +1,9 @@
 // @ts-ignore
-import * as stdFs from "https://deno.land/std@0.89.0/fs/mod.ts";
+import { ensureDir, ensureDirSync } from "https://deno.land/std@0.104.0/fs/ensure_dir.ts";
 // @ts-ignore
-import * as stdPath from "https://deno.land/std@0.89.0/path/mod.ts";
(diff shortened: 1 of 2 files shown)


packages/common/src/runtimes/DenoRuntime.ts
```diff
@@ -1,7 +1,9 @@
 // @ts-ignore
-import * as stdFs from "https://deno.land/std@0.89.0/fs/mod.ts";
+import { ensureDir, ensureDirSync } from "https://deno.land/std@0.104.0/fs/expand_glob.ts";
 // @ts-ignore
-import * as stdPath from "https://deno.land/std@0.89.0/path/mod.ts";
+import { expandGlob, expandGlobSync } from "https://deno.land/std@0.104.0/fs/ensure_dir.ts";
```
dsherret: Oops. The module specifiers are backwards.

dsherret (commented):
Thanks for this PR! I've been meaning to do it, but have been lazy.

packages/common/src/runtimes/DenoRuntime.ts
wojpawlik: Fixed.

dsherret:
Thanks @wojpawlik! I'll do a release later.

PR #1241, opened 2022-01-21
Fix jsx element duplicate leading trivia

### LINKED ISSUE #1240: Bug: JsxOpeningElement | JsxSelfClosingElement addAttribute with leadingTrivia results in duplicate comments
**Describe the bug**

Version: 13.0.2

When attempting to add new attributes with comments to a `JsxOpeningElement` or `JsxSelfClosingElement` node (using the `leadingTrivia` property in the structure), it seems to add two of the same comment.

**To Reproduce**

```ts

const sourceFile = project.createSourceFile("file.tsx",
    `
        const Example = () => {
            return (
                <Button>
                    Click Me
                </Button>
            );
        }
    `
);

const buttonJsxElement = sourceFile.getFirstDescendantByKindOrThrow(
    SyntaxKind.JsxOpeningElement
);

buttonJsxElement.addAttribute({
    name: "onClick",
    initializer: "{() => {}}",
    leadingTrivia: "// This handles the click"
});
```

I've also spun up a CodeSandbox that demonstrates the behavior with `JsxSelfClosingElement` too: https://codesandbox.io/s/ts-morph-addattribute-with-leadingtrivia-adds-duplicate-comments-p53ex?file=/src/create-source-file.ts

**Expected behavior**

I would expect the `leadingTrivia` to appear once above the attribute, i.e. 

```tsx
<Button
    // This handles the click
    onClick={() => {}}>
    Click Me
</Button>
```

**Actual behavior**

The `leadingTrivia` is added twice.

```tsx
<Button // This handles the click
    // This handles the click
    onClick={() => {}}>
    Click Me
</Button>
```

brandongregoryscott:
Resolves #1240

After some stack tracing in a test with the expected output, it appears that `JsxAttributeDeciderStructurePrinter` was calling `printText`, which eventually calls down into the base `NodePrinter` which already handles printing the leading/trailing trivia. I think there was some sort of accidental recursion/duplicate `printText`/`printTextInternal` calls firing off throughout the inheritance hierarchy resulting in the duplicated trivia. 

Additionally, I added tests for the `JsxSpreadAttribute` codepath - I noticed something odd. The `JsxSpreadAttributeStructure` has an `expression` along with the base structure properties `leadingTrivia`/`trailingTrivia`, but the writer essentially prepends `...` without brackets.

Unless I'm totally missing something, this is invalid JSX: `<div ...props />` (should be `<div {...props} />`). I wrote the test assertion to match the current behavior, but I think that writer might need to be updated too.

### DIFF
--- a/packages/ts-morph/src/factories/StructurePrinterFactory.ts
+++ b/packages/ts-morph/src/factories/StructurePrinterFactory.ts
@@ -1,7 +1,7 @@
 // DO NOT EDIT - Automatically maintained by createStructurePrinterFactory.ts
 import { Memoize } from "@ts-morph/common";
-import * as structurePrinters from "../structurePrinters";
 import { SupportedFormatCodeSettings } from "../options";
+import * as structurePrinters from "../structurePrinters";
 
 /** Cached lazy factory for StructurePrinters. */
 export class StructurePrinterFactory {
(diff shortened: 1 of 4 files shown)


dsherret:
Thanks @brandongregoryscott!

> Unless I'm totally missing something, this is invalid JSX: <div ...props /> (should be <div {...props} />). I wrote the test assertion to match the current behavior, but I think that writer might need to be updated too.

Yes, you're right. I will look into that now and then do a new release.

PR #1274, opened 2022-05-13
Remove duplicated test file with uppercase name

jakebailey:
This file is a duplicate of `mappedTypeNodeTests.ts`; the existence of both in the repo makes git unhappy when checking out on Windows/macOS which have case-insensitive paths.


dsherret (approved):
Thanks! I noticed this the other day when I checked out the repo on my laptop and was too lazy to fix it.

PR #1275, opened 2022-05-13
Fix undefined in comment node type guards

jakebailey:
I hit this while working on a transform; the `as` hides the fact that these can be undefined thanks to `node?`. Maybe I should change the cast too.

I'm not sure if I'm sending PRs correctly; there are a lot more changes in the repo when I build it, but they don't look related to any change I've made at all. Happy to let you make this change instead if I'm doing it wrong!


dsherret (approved):
LGTM. Thanks!

Yeah, repo is a bit of a mess and slow to work in atm because some things are in flux. I'll be rewriting it to be deno first in the future and compile for npm with dnt (https://github.com/denoland/dnt) in order to get rid of the `deno` directory.

PR #1288, opened 2022-06-03
feat: add Type.isTemplateLiteral method

shaharkazaz:
Expose a new `isTemplateLiteral` method from the `Type` class.

No breaking changes were introduced in this PR.


docs/details/types.md
```diff
@@ -200,25 +200,26 @@ type.isAnonymous();
 type.isAny();
```
shaharkazaz, line 200: I ordered them alphabetically and added `isTemplateLiteral` to the list, that's why you see multiple changes here 🙂

dsherret (approved):
Also looks good. Thanks!

packages/ts-morph/src/compiler/types/Type.ts
```diff
@@ -389,6 +389,13 @@ export class Type<TType extends ts.Type = ts.Type> {
     return symbol.getName() === "Array" && this.getTypeArguments().length === 1;
   }
 
+    /**
+     * Gets if this is a boolean type.
```
dsherret, line 393: template literal. I'll fix it after merge (going to do a release)

shaharkazaz: Missed that 😅

PR #1327, opened 2022-09-09
Feature/error messages

### LINKED ISSUE #715: Providing custom messages to OrThrow methods
**Is your feature request related to a problem? Please describe.**
It's currently not possible to provide specific error messages to the `XXXOrThrow` methods. This limits these "AST assertions" to generic error message that most likely won't help a user of your library.

To put it another way: the current error messages can only tell you *what* was expected, not *why* it was expected.

**Describe the solution you'd like**
I'd love to add an optional `message` parameter to all the `OrThrow` methods. If this parameter isn't specified the current 'generic' messages should be used.

**Describe alternatives you've considered**
You could try-catch the assertions and throw your own error but this is clunky and doesn't really work for multiple assertions.

jantimon:
This is a POC branch to improve error messages in two ways:

1. Allow custom error messages #715
2. Show the source location of the error 

error

To finish this PR I would need feedback on the approach itself and your help to improve `errors.ts`.

Closes #715


packages/common/src/errors.ts
```diff
@@ -2,43 +2,74 @@ import { StandardizedFilePath } from "./fileSystem";
...
+          const brokenLineEnd = sourceCode.indexOf("\n", pos);
+          const brokenLine = sourceCode.substring(brokenLineStart + 1, brokenLineEnd === -1 ? undefined : brokenLineEnd);
+          source = { fileName: sourceFile.getFilePath(), pos: { line, character: pos - brokenLineStart } };
+          const linePrefix = `> ${source.pos.line + 1} |`;
+          messageWithSource += `\nin ${source.fileName}:${source.pos.line + 1}:${source.pos.character + 1}\n\n${linePrefix}${brokenLine}\n${" ".repeat(linePrefix.length - 1)}|${" ".repeat(source.pos.character)}^`;
+        } catch (e) {
+          // Errors inside BaseError would be confusing
+          // so ignore errors here and fallback to the original message
+        }
+    }
```
dsherret: Would be good if we added some tests for this.

dsherret (changes requested):
@jantimon sorry for my delay! This looks really good. Thanks! Would you be able to add a few tests for the more complicated functionality where it prints out the message? Then we should be good to merge this.

deno/common/ts_morph_common.d.ts
```diff
@@ -345,28 +345,41 @@ export interface TsSourceFileContainer {
 /** Decorator for memoizing the result of a method or get accessor. */
 export declare function Memoize(target: any, propertyName: string, descriptor: TypedPropertyDescriptor<any>): void;
 
+/** Minimal attributes to show a error message with source */
+export declare type MaybeTraceAbleNode = undefined | null | {} | {
```
dsherret: Maybe just undefined?

```suggestion
export declare type MaybeTraceAbleNode = undefined | {
```

packages/ts-morph/src/Project.ts
```diff
@@ -242,10 +242,10 @@ export class Project {
    * Gets a directory by the specified path or throws if it doesn't exist.
    * @param dirPath - Path to create the directory at.
    */
-  getDirectoryOrThrow(dirPath: string): Directory {
+  getDirectoryOrThrow(dirPath: string, message?: string): Directory {
```
dsherret: How hard would it be to make these `onErrorMessage?: string | () => string` so someone could provide a string that gets lazily created only when necessary?

jantimon:
no worries - I am happy you like the idea :)

packages/ts-morph/src/Project.ts
jantimon: Good idea - I will take a look 👍

packages/common/src/errors.ts
jantimon: Added some tests :)  
I am not sure how to setup the test inside `packages/common` so I had to place the test inside `packages/ts-morph`

deno/common/ts_morph_common.d.ts
jantimon: Done 👍

Is there any way to use `Node` type from `ts-morph` inside the `common` package?

packages/ts-morph/src/Project.ts
jantimon: Done

jantimon:
@dsherret if it helps you to save some time on this issue we can also discuss this PR in a video call

dsherret (approved):
Did a few changes and it LGTM now. Thanks for this @jantimon!

jantimon:
wow nice! :)

PR #1371, opened 2022-12-13
Add #remove on JsxElement and JsxSelfClosingElement

dherault:
Add support for JsxElement.remove and JsxSelfClosingElement.remove

If anything just ask I'm around.

### DIFF
--- a/packages/ts-morph/src/tests/compiler/ast/jsx/jsxElementTests.ts
+++ b/packages/ts-morph/src/tests/compiler/ast/jsx/jsxElementTests.ts
@@ -1,6 +1,6 @@
 import { errors, nameof, SyntaxKind } from "@ts-morph/common";
 import { expect } from "chai";
-import { JsxElement } from "../../../../compiler";
+import { JsxElement, JsxSelfClosingElement } from "../../../../compiler";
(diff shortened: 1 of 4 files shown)


packages/ts-morph/src/tests/compiler/ast/jsx/jsxElementTests.ts
```diff
@@ -159,4 +159,34 @@ const v = <div>
...
+      }
+      catch (err) {
+        error = err;
+      }
+
+      expect(error).to.be.instanceOf(errors.InvalidOperationError);
+    });
+
+    it("should remove the JsxElement child", () => {
+      doTestWithJsxElementChild(`var t = (<jsx><jsx2></jsx2></jsx>);`, `var t = (<jsx></jsx>);`);
```
dsherret, line 189: Thanks, but it's not so simple unfortunately. Once spaces are added then it gets much more complicated because spaces have meaning in JSX. It's why I haven't had the time to add this functionality yet. For example:

```ts
it("should remove the JsxElement child when has spaces", () => {
      doTestWithJsxElementChild(
        `var t = (<jsx>
        <jsx2></jsx2>
      </jsx>);`,
        `var t = (<jsx></jsx>);`,
      );
    });
```

PR #1386, opened 2023-03-03
docs: update transforms example to use factory

tobyhede:
ts.createNumericLiteral is deprecated, should use factory instead.


dsherret (approved):
Thanks!

PR #1397, opened 2023-03-28
feat: add JSDocable to ExportAssignment

giuseppelt:
export assignment can define jsdoc

```ts
/** basic assignment */
export = 1;

const hello = "";
/** default assignment */
export default hello;
```


dsherret (approved):
Thanks!

PR #1408, opened 2023-05-10
feat: improve typing of `Type`

odiak:
I made some methods type guard.


dsherret (approved):
Thanks! This is a nice change.

Issue #1409, opened 2023-05-10
Replacing JsxSelfClosingElement with text failed


odiak:
**Describe the bug**

Version: 18.0.0

I tried to replace a JsxSelfClosingElement with text, but it failed with following error:

<details>
<summary>error log</summary>

```
/project/node_modules/ts-morph/dist/ts-morph.js:39
            throw new common.errors.InvalidOperationError("Cannot get the next when at the end of the iterator.");
                  ^
ManipulationError: Manipulation error: Cannot get the next when at the end of the iterator.

-- Details --
Path: /project/foo.tsx
Text: "const F = () => {\n    return <div>\n      hello\n      world\n    </div>\n}\n"
Stack: Error: Cannot get the next when at the end of the iterator.
    at AdvancedIterator.next (/project/node_modules/ts-morph/dist/ts-morph.js:39:19)
    at RangeParentHandler.handleNode (/project/node_modules/ts-morph/dist/ts-morph.js:1683:78)
    at ParentFinderReplacementNodeHandler.handleNode (/project/node_modules/ts-morph/dist/ts-morph.js:1567:36)
    at NodeHandlerHelper.handleForValues (/project/node_modules/ts-morph/dist/ts-morph.js:1373:21)
    at ParentFinderReplacementNodeHandler.handleChildren (/project/node_modules/ts-morph/dist/ts-morph.js:1440:25)
    at ParentFinderReplacementNodeHandler.handleNode (/project/node_modules/ts-morph/dist/ts-morph.js:1430:18)
    at ParentFinderReplacementNodeHandler.handleNode (/project/node_modules/ts-morph/dist/ts-morph.js:1570:19)
    at NodeHandlerHelper.handleForValues (/project/node_modules/ts-morph/dist/ts-morph.js:1373:21)
    at ParentFinderReplacementNodeHandler.handleChildren (/project/node_modules/ts-morph/dist/ts-morph.js:1440:25)
    at ParentFinderReplacementNodeHandler.handleNode (/project/node_modules/ts-morph/dist/ts-morph.js:1430:18)
    at throwError (/project/node_modules/ts-morph/dist/ts-morph.js:2298:19)
    at doManipulation (/project/node_modules/ts-morph/dist/ts-morph.js:2296:9)
    at insertIntoParentTextRange (/project/node_modules/ts-morph/dist/ts-morph.js:2317:5)
[...]

PR #1410, opened 2023-05-10
fix: fix manipulation of JsxSelfClosingElement

odiak:
While I investigating #1409, I found suspicious code from stack trace.
Because it works as expected after fixing there, I made this pull request.

### DIFF
--- a/packages/ts-morph/src/manipulation/nodeHandlers/RangeHandler.ts
+++ b/packages/ts-morph/src/manipulation/nodeHandlers/RangeHandler.ts
@@ -56,7 +56,7 @@ export class RangeHandler implements NodeHandler {
       newNodeChildren.next();
 
     // handle the rest
-    while (!currentNodeChildren.done)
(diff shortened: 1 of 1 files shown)


odiak:
I also found some similar code pieces, though I'm not sure if they cause error.

https://github.com/dsherret/ts-morph/blob/c12673319f919a4386685c63d31356cd2c1a981c/packages/ts-morph/src/manipulation/nodeHandlers/RangeParentHandler.ts#L89-L90

https://github.com/dsherret/ts-morph/blob/c12673319f919a4386685c63d31356cd2c1a981c/packages/ts-morph/src/manipulation/nodeHandlers/DefaultParentHandler.ts#L64-L65

https://github.com/dsherret/ts-morph/blob/c12673319f919a4386685c63d31356cd2c1a981c/packages/ts-morph/src/manipulation/nodeHandlers/UnwrapParentHandler.ts#L50-L51

PR #1412, opened 2023-05-12
fix: allow writer functions in property named structures

### LINKED ISSUE #1160: Support automatically determining if a string literal should be written for a property key or value in Writers.object
**Describe the bug**

Version: 10.1.0

I want to generate .ts file with variable content, that depend on my `existedObject`
`Writers` generate incorrect object representation

**To Reproduce**

```ts
import { Project, VariableDeclarationKind, Writers } from "ts-morph";

const sourceFile = project.createSourceFile("test.ts", ``);

const existedObject = {
      id: "some-id",
      "quote-key": 2
}

sourceFile.addVariableStatement({
  isExported: true,
  declarationKind: VariableDeclarationKind.Const,
  declarations: [{
    name: 'flavor',
    initializer: Writers.object(existedObject),
  }]
})

console.log(sourceFile.getFullText())
```

Output:
```
export const flavor = { 
        id: some-id, 
        quote-key: 2 
    }; 
```

**Expected behavior**

```
export const flavor = { 
        id: "some-id", 
        "quote-key": 2 
    }; 
```

forivall:
Fixes #1160

Other solutions possible, as mentioned in #1160 comments


packages/ts-morph/src/structures/base/name/PropertyNamedNodeStructure.ts
```diff
@@ -1,3 +1,5 @@
+import { WriterFunction } from "../../../types";
+
 export interface PropertyNamedNodeStructure {
-  name: string;
+  name: string | WriterFunction;
```
dsherret, line 4: I'm hesitant to merge this because people are likely doing comparisons on `string`. One of the major issues with structures is that changes that make it easier for writing can make it harder to use for reading. At this point, I think this might be too breaking.

PR #1417, opened 2023-06-13
fix: allow `.tranform` to work with Nodes from another parsed source file

### LINKED ISSUE #1416: `transform` looses `StringLiteral` nodes
**Describe the bug**

Version: 18.0.0

I want to replace a node with another node. As far as I understand thsi should be done using the `transform` method. It works as expected except when the replacement node contains `StringLiteral`s. In that case those literals get lost and are missing from the output.

**To Reproduce**

Replace a node containing a `StringLiteral` using the `transform` method.

```ts
import { Project, Node } from "ts-morph"

const printCode = (node: Node) => node.print().trim()

const parseCode = (code: string) => new Project()
	.createSourceFile("test.ts", code, { overwrite: true })
	.getStatement(Node.isVariableStatement)
	?.getDeclarationList()
	.getDeclarations()[0]
	.getInitializer()!

const toReplace = parseCode('const x = 1')
const parent = toReplace.getSourceFile()!

const replaceWith = parseCode('const y = console.log("test")')
toReplace.transform(() => replaceWith.compilerNode)

console.log('expected:    ', 'const x = console.log("test")');
console.log('actual:      ', printCode(parent));
```

The output of the code is:
```
expected:    const x = console.log("test")
actual:      const x = console.log()
```

**Expected behavior**

I would expect `transform` to replace the node with all it's children correctly.

ivanhofer:
closes #1416


ivanhofer:
A test fails. Somehow the changed code breaks something with JSDoc comments. Not sure how to fix this.
The fix sometimes also outputs the code with another formatting.

packages/ts-morph/src/compiler/ast/common/Node.ts
```diff
@@ -1565,7 +1565,7 @@ export class Node<NodeType extends ts.Node = ts.Node> {
 
         for (const transform of transformations) {
           finalText += fileText.substring(lastPos, transform.start);
-          finalText += printer.printNode(ts.EmitHint.Unspecified, transform.compilerNode, compilerSourceFile);
+          finalText += printer.printNode(ts.EmitHint.Unspecified, transform.compilerNode, transform.compilerNode.getSourceFile());
```
dsherret:
```suggestion
          finalText += printer.printNode(ts.EmitHint.Unspecified, transform.compilerNode, transform.compilerNode.getSourceFile() ?? compilerSourceFile);
```

It probably needs to be this.

dsherret (changes requested):
This seems pretty reasonable to do actually. Just needs to fallback to using the initial source file I think.

packages/ts-morph/src/compiler/ast/common/Node.ts
dsherret: Actually, it looks like the initial code was wrong...

dsherret: No, ok I remember it now. This suggestion was correct.

dsherret (approved):
LGTM. Thanks!

PR #1436, opened 2023-08-07
fix: support dynamic type imports in getReferencedSourceFiles()

### LINKED ISSUE #1435: getReferencedSourceFiles() doesn’t resolve dynamic imports
I want to analyze the dependency tree of typescript files using `getReferencedSourceFiles()`. I’ve noticed that it doesn’t work for file with dynamic imports, as for example:

```typescript
declare const foo: import("./foo").Foo
```

When enabling logging, I get the following warning: `Unknown import string literal parent: LiteralType`.

I get a valid result for the dynamic import when calling `getImportStringLiterals()`. So I assume that the `ImportType` Node needs to be handled somehow in [`getSourceFileForLiteral()`](https://github.com/dsherret/ts-morph/blob/382f0c8a9d53326bc295111d722973bbd428d805/packages/ts-morph/src/utils/references/SourceFileReferenceContainer.ts#L79C11-L79C34).

lo1tuma:
Fixes #1435


dsherret (approved):
Thanks!

PR #1451, opened 2023-09-11
feat: automatically quote property keys if necessary when writing structures

### LINKED ISSUE #1160: Support automatically determining if a string literal should be written for a property key or value in Writers.object

dsherret:
Closes #1160


dsherret:
Reverted in https://github.com/dsherret/ts-morph/commit/055bf2158539d0b32c71141ca7c5e5213b85a611 because this will break computed property names.

PR #1456, opened 2023-09-21
feat: add wrapped method for `getShorthandAssignmentValueSymbol` to `TypeChecker`. add convenience method for getting the value symbol on `ShorthandPropertyAssignment`.

kronodeus:
Issue: https://github.com/dsherret/ts-morph/issues/1457

All tests passing. Should carry no regression risk as it is a net-new API. See issue for usage example.


dsherret (approved):
LGTM. Thanks!

PR #1460, opened 2023-10-10
feat: `ProjectOptions::defaultCompilerOptions`

### LINKED ISSUE #1459: Override default `compilerOptions`
How can I make deno2node default to `"strict": true`, while allowing to override the option via `tsconfig.json` (or CLI flags)?

wojpawlik:
Closes #1459.


dsherret (approved):
Seems reasonable. Thanks!

Issue #1494, opened 2024-01-26
Add support for module specifier "aliases" (tsconfig.paths mappings) when moving files/folders.


tryggvigy:
**tl;dr** Add support for module specifier "aliases" (tsconfig.paths mappings) when moving files/folders.

---

Let's say your tsconfig contains a paths mapping like this:
```
    "paths": {
      "Root/*": ["./*"]
    },
```

I noticed that 

```ts
project.createDirectory('someDir');
const srcFile = project.getSourceFileOrThrow('someFile.ts');
srcFile.moveToDirectory('someDir');
```

will update all relative import/export specifiers referencing 'someFile.ts' but _not_ if the specifier used an alias such as 

```ts
import * from 'Root/someFile';
```

I expected `ts-morph` to update the specifier to `import * from 'Root/someDir/someFile';` but it doesn't.
I think it would be nice feature to support this. VSCode can handle paths mapped import/exports when moving files.

**Describe the solution you'd like**

I think the problem is that [`updateStringLiteralReferences(nodeReferences: ReadonlyArray<[StringLiteral, SourceFile]>) {`](https://github.com/dsherret/ts-morph/blob/930779a0d3699be35ab17bbe9dd26ec00350fe53/packages/ts-morph/src/compiler/ast/module/SourceFile.ts#L1025C1-L1028C4) only updates the module specifier if it's a relative path like `'../someFile'` not if it's an module specifier alias like `'Root/someFile'`.

PR #1495, opened 2024-01-28
Support tsConfig paths aliases during move operations

tryggvigy:
Showcases fix for https://github.com/dsherret/ts-morph/issues/1494.

Naive support for `tsconfig.compilerOptions.paths` module specifier aliases during move operations.

```ts
const project = new Project({
  useInMemoryFileSystem: true,
  compilerOptions: {
    paths: {
      'Root/*': ['./*']
    }
  }
});

const a = project.createSourceFile("a.ts", `export function foo() {}`);
// b.ts imports `a` via paths mapping alias.
const b = project.createSourceFile("b.ts", `import { foo } from "Root/a";`);

// Move ./a.ts -> ./a./a.ts
const dir = project.createDirectory('./a');
a.moveToDirectory(dir);

// Updating path aliased import works.
expect(b.getFullText()).to.equal(`import { foo } from "Root/a/a";`);
```

### DIFF
--- a/packages/ts-morph/src/compiler/ast/module/SourceFile.ts
+++ b/packages/ts-morph/src/compiler/ast/module/SourceFile.ts
@@ -1023,7 +1023,26 @@ export class SourceFile extends SourceFileBase<ts.SourceFile> {
 
 function updateStringLiteralReferences(nodeReferences: ReadonlyArray<[StringLiteral, SourceFile]>) {
   for (const [stringLiteral, sourceFile] of nodeReferences) {
-    if (ModuleUtils.isModuleSpecifierRelative(stringLiteral.getLiteralText()))
+    const compilerOptions = stringLiteral._context.compilerOptions.get()
+    const literalText = stringLiteral.getLiteralText();
+    const paths = compilerOptions?.paths;
+    // See https://www.typescriptlang.org/docs/handbook/modules/reference.html#paths
+    if (paths) {
+      // Check if the string literal is a path alias specifier
+      for (const [path, mappings] of Object.entries(paths)) {
+        const hasWildCard = path.endsWith('*');
+        const [alias] = path.split('*');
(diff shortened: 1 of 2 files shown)


packages/ts-morph/src/compiler/ast/module/SourceFile.ts
```diff
@@ -1023,7 +1023,26 @@ export class SourceFile extends SourceFileBase<ts.SourceFile> {
...
+        if (literalText.startsWith(alias) && hasWildCard) {
+          const pathAfterAlias = FileUtils.pathJoin(
+            stringLiteral._sourceFile.getDirectoryPath(),
+            stringLiteral._sourceFile.getRelativePathAsModuleSpecifierTo(sourceFile)
+          );
+          stringLiteral.setLiteralValue(FileUtils.pathJoin(alias, pathAfterAlias));
+          break;
+        }
+      }
+    }
```
tryggvigy, line 1044: This is very naive right now and only works with wildcard paths (most common use-case afaik). If we want to productionize this approach we should create new helpers in `ModuleUtils` and/or `SourceFile` to more deeply support paths aliases, instead of reaching directly for FileUtils here.

packages/ts-morph/src/tests/issues/1494tests.ts
```diff
@@ -0,0 +1,42 @@
...
+import { Project } from "../../Project";
+
+describe("tests for my issue", () => {
+  it("should update specifiers with path aliases during move", () => {
+    const project = new Project({
+      useInMemoryFileSystem: true,
+      compilerOptions: {
+        baseUrl: '.',
+        paths: {
+          'Root/*': ['./*']
```
tryggvigy, line 12: If we want to move ahead with this PR we should add a lot more tests for paths.

- With and without wildcard value
- More nested mappings

PR #1549, opened 2024-06-26
perf: remove mkdirp dependency

benmccann:


dsherret (approved):
Thanks!

dsherret (changes requested):
Seems to have introduced a regression. See CI output. We'll just have to update this code to not throw for existing directories.

benmccann:
sorry, I can't figure out how to run the project locally. I've never used Deno before and am not familiar with what's happening in the build process. I noticed an error in the code and updated it, so hopefully that will get the tests passing. If not, I may need a little help with this one

dsherret:
I think it should just be to run `npm install` in the root, then `cd packages/common`, then `npm run test`

benmccann:
ah, cool. thanks. that worked! I was trying to build it from the root first, which wasn't working at all

benmccann:
@dsherret would you be able to give this PR another peek now that the CI is passing? thanks!

dsherret (approved):
LGTM. Thanks!

PR #1557, opened 2024-07-18
fix: getDerivedClasses() isn't correct in some cases

### LINKED ISSUE #1556: getDerivedClasses() isn't correct in some cases
**Describe the bug**

Version: 23.0.0

If a class extends another class that is in a PropertyAccessExpression, `getDerivedClasses()` will not register the subclass as a derived class. To make this more concrete, I'll give an example below.

I have an exported class `ExampleClass` in source-file `example-module.ts` and I `import * as example_module from "./example-module";` in another module. Now I have another class `ExampleSubclass` that extends `example_module.ExampleClass". Then a call to `getDerivedClasses()` on `ExampleClass`'s `ClassDeclaration` node will not include `ExampleSubclass`'s `ClassDeclaration`.

I do have a fix that is a 1 line change that I will submit a PR for soon.

**To Reproduce**

```ts

const sourceFile1 = project.createSourceFile("test.ts", `export class ExampleClass {}`);
const sourceFile2 = project.createSourceFile("test2.ts", `import * as test from "./test"; class ExampleSubclass extends test.ExampleClass {};`);

// reproduce the problem here
const derivedClasses = sourceFile1.getClassOrThrow("ExampleClass").getDerivedClasses();
console.log(derivedClasses.length);
```

**Expected behavior**

1 should be logged since `ExampleClass` has 1 derived class. Instead, 0 is logged.

jmyrick02:
Fixes #1556 

I'm unsure if the test is desirable since it imports `Project`. I wasn't sure how to test the fix without creating a `Project`. Let me know.

### DIFF
--- a/packages/ts-morph/src/compiler/ast/class/base/ClassLikeDeclarationBase.ts
+++ b/packages/ts-morph/src/compiler/ast/class/base/ClassLikeDeclarationBase.ts
@@ -1241,8 +1241,9 @@ function getImmediateDerivedClasses(classDec: ClassLikeDeclarationBaseSpecific &
   if (nameNode == null)
     return classes;
 
-  for (const node of nameNode.findReferencesAsNodes()) {
-    const nodeParent = node.getFirstAncestorByKind(SyntaxKind.ExpressionWithTypeArguments);
+  for (let node of nameNode.findReferencesAsNodes()) {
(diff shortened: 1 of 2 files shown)


dsherret (approved):
Thanks!

PR #1558, opened 2024-07-22
perf: switch to tinyglobby to drop 15 dependencies

benmccann:
https://npmgraph.js.org/?q=fast-glob - 17 dependencies
https://npmgraph.js.org/?q=tinyglobby - 2 dependencies

I think the even better solution though would be to move globbing into user land. The users who want globbing can invoke it on their own and pass the result into `addSourceFilesAtPaths` so that users who don't need globbing don't have to pay that cost.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1656,12 +1656,12 @@
       }
     },
     "node_modules/tinyglobby": {
-      "version": "0.1.2",
-      "resolved": "https://registry.npmjs.org/tinyglobby/-/tinyglobby-0.1.2.tgz",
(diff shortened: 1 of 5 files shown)


benmccann:
@dsherret any chance you might be able to take a look at this? thanks!

benmccann:
@dsherret thanks for the merge of https://github.com/dsherret/ts-morph/pull/1549. I've rebased this PR and updated it so that it's passing the CI now

dsherret (approved):
Thanks!

> I think the even better solution though would be to move globbing into user land.

Ship has sailed.

dsherret:
Going to revert this one in https://github.com/dsherret/ts-morph/pull/1595 -- Maybe in a few years once things have played out with tinyglobby I'll switch back to that.

PR #1560, opened 2024-07-25
fix: getBodyText() doesn't work as would be expected

### LINKED ISSUE #1559: getBodyText() doesn't work as would be expected
**Describe the bug**

There are two issues I'd like to discuss here. One is objectively a bug, while the other is more opinionated.

The first issue (the definite bug) is how `getBodyTextWithoutLeadingIndentation()` handles a line that only contains indentation (spaces/tabs). It will cause bad output as I will outline below.

The second issue (more opinionated) is that empty lines in a method body cause `getBodyText()` to not remove any indentation at all. This is how I found this issue, since I'm trying to print method bodies at a standard indentation level, but whenever the body contains an empty line, it causes the method to be displayed much more indented than otherwise. I'll give an example below.

I do have a fix for these two issues working locally with tests, so I'll contribute those in a PR after submitting this issue.

Version: 23.0.0

**To Reproduce**

In the section Issue 1, the output log statement has the two `console.log()` statements at differing indentation levels, when they should be at the same indentation level. I have a fix for this issue that I will submit a PR for.

The section Issue 2, the output is the same, but I would prefer empty lines to be ignored. I think this makes sense since a single empty line in a method body shouldn't affect the whole indentation level of the body text in my opinion. I also have a fix for this behavior in the PR I will submit.

```ts

// Issue #1

let sourceFile = project.createSourceFile("test.ts", `function foo() {\n  console.log()\n  \n  console.log()\n}`);

// reproduce the problem here
let foo = sourceFile.getFunction("foo");
// we expect the two console.log() statements to be at the same indentation level
console.log(foo?.getBodyText());

// Issue #2
sourceFile = project.createSourceFile("test2.ts", `function foo() {\n  console.log()\n\n  console.log()\n}`);

// reproduce the problem here
foo = sourceFile.getFunction("foo");
// my preference for behavior would be for the new line without any indentation to not affect the output and remove the indentation as if the line didn't exist in the method
console.log(foo?.getBodyText());
```

**Expected behavior**

In case 1, I expect the two console log statements to have the same indentation level, that of 0 since the minimum indentation level in the method body is 1 and all lines have that indentation level.

In case 2, I would prefer an empty line to be ignored, though this is more opinionated, but I think it makes sense for use cases I can imagine, and especially for my use case.

jmyrick02:
Fixes #1559 

The first test corresponds to the first issue mentioned in #1559, which is definitely a bug. This is addressed by the addition of the new if statement at the end of the while loop. The second test corresponds to my preferred behavior of ignoring empty lines when adjusting indentation levels. This was addressed by the modification of the condition to update the minimum indent level. If you do not believe this should be default behavior, at least an option should be provided, since I'm finding it necessary for my use case.

Please let me know if you have any thoughts.

### DIFF
--- a/packages/common/src/tests/utils/stringUtilsTests.ts
+++ b/packages/common/src/tests/utils/stringUtilsTests.ts
@@ -331,6 +331,14 @@ describe("StringUtils", () => {
       str += "\n    other";
       doTest(str, "this is a `\n    test`\nother", { isInStringAtPos: index => index >= pos && index < end });
     });
+
+    it("should handle lines that only have indentation", () => {
+      doTest("  test\n  \n  test", "test\n\ntest", { indentSizeInSpaces: 2 });
+    });
+
+    it("should ignore empty lines", () => {
+      doTest("  test\n\n  test", "test\n\ntest", { indentSizeInSpaces: 2 });
+    });
   });
 
   describe(nameof(StringUtils, "indent"), () => {
--- a/packages/common/src/utils/StringUtils.ts
+++ b/packages/common/src/utils/StringUtils.ts
@@ -183,11 +183,15 @@ export class StringUtils {
 
         // indentation for spaces rounds up to the nearest tab size multiple
         const indentWidth = Math.ceil(spacesCount / indentSizeInSpaces) * indentSizeInSpaces + tabsCount * indentSizeInSpaces;
-        if (minIndentWidth == null || indentWidth < minIndentWidth)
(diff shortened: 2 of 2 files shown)


dsherret (approved):
Thanks!

jmyrick02:
> Thanks!

Thank you so much! This is my first real OSS contribution, so I appreciate this a lot. Thanks for being a maintainer.

Issue #1586, opened 2024-11-22
NotImplementedError: Unexpected! Inserting syntax kind of ExportAssignment, but FunctionDeclaration was inserted.


bennycode:
**Describe the bug**

Version: 24.0.0

I cannot default export a function declaration.

**To Reproduce**

The following default export is allowed in JavaScript/TypeScript ([see here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export#syntax)):

```ts
export default function functionName() {}
```  

I want to insert such default export with a function declaration using `insertExportAssignment`:

```ts
sourceFile.insertExportAssignment(position, {
  expression: 'function functionName() {}',
  isExportEquals: false,
});
```

Unfortunately, it makes my code crash:

> NotImplementedError: Unexpected! Inserting syntax kind of ExportAssignment, but FunctionDeclaration was inserted.

I have to work around this error by using:

```ts
sourceFile.insertExportAssignment(position, {
  expression: 'functionName() {}',
  isExportEquals: false,
});
``` 

Or by doing:

```ts
sourceFile.insertStatements(position, 'export default function functionName() {}');
```
**Expected behavior**

Not crashing when using:

```ts
sourceFile.insertExportAssignment(position, {
  expression: 'function functionName() {}',
  isExportEquals: false,
});
```

dsherret:
In this case, you'll want to use `addFunction` with `isDefaultExport: true`.

Function declarations with a default export and export assignments are represented differently in the typescript compiler:

1. https://ts-ast-viewer.com/#code/KYDwDg9gTgLgBAE2AMwIYFcA28bAM4wDcQA
2. https://ts-ast-viewer.com/#code/KYDwDg9gTgLgBAE2AMwIYFcA29noHYDGMAlhHnKKgLZibAAUAlHAN4C+QA

PR #1587, opened 2024-11-22
feat: allow exporting function declarations

bennycode:
I would like to solve #1586 and allow exporting a function declaration with a `default export` statement. I added a test case to see what happens if I pass the expression `functionName() {}` to `insertExportAssignments`. It actually creates this:

```ts
export default functionName() {};
``` 

This result is invalid according to the TS Playground: https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBAE2AMwIYFcA29noHYDGMAlhHgHKoC2wAFAJRwDecAvgNxA

@dsherret Shouldn't the output be `export default function functionName() {};`?

### DIFF
--- a/packages/ts-morph/src/tests/compiler/ast/base/moduledNodeTests.ts
+++ b/packages/ts-morph/src/tests/compiler/ast/base/moduledNodeTests.ts
@@ -235,7 +235,7 @@ describe("ModuledNode", () => {
       expect(sourceFile.getFullText()).to.equal(expectedCode);
     }
 
-    it("should insert the different kinds of exports", () => {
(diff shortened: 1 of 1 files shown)


dsherret:
Thanks, but see https://github.com/dsherret/ts-morph/issues/1586#issuecomment-2494304753

Note the return type of `insertExportAssignment` is `ExportAssignment`, but `export default function name()` is a `FunctionDeclaration`.

bennycode:
Thanks for checking this! Shouldn't the test case fail for the input of `{ isExportEquals: false, expression: "functionName() {}" }` as it produces invalid TypeScript code?

PR #1592, opened 2024-12-10
Fix: ts-morph incorrectly transforms nodes which begin with whitespace

### LINKED ISSUE #1591: ts-morph incorrectly transforms nodes which begin with whitespace
if you have a node such as:

```
<Text><Text>Foo</Text> is great</Text>
```

and you do a transform:

```
node.transform((traversal) => {
    if (ts.isJsxText(traversal.currentNode) && traversal.currentNode.text != 'Foo'){
        return traversal.factory.createJsxText(' is NOT great', false);
    } else {
        return traversal.visitChildren();
    }
});
```

the result should be:

```
<Text><Text>Foo</Text> is NOT great</Text>
```

but it is:

```
<Text><Text>Foo</Text>  is NOT great</Text>
```

instead. Notice how an extra space was added. This is due to the following error in the code:

```
        function handleTransformation(oldNode, newNode) {
            if (oldNode === newNode && newNode.emitNode == null)
                return;
            const start = oldNode.getStart(compilerSourceFile, true); //This is the error: start should be oldNode.pos which correctly accounts for whitespace
            const end = oldNode.end;
            let lastTransformation;
            while ((lastTransformation = transformations[transformations.length - 1]) && lastTransformation.start > start)
                transformations.pop();
            const wrappedNode = compilerFactory.getExistingNodeFromCompilerNode(oldNode);
            transformations.push({
                start,
                end,
                compilerNode: newNode,
            });
            if (wrappedNode != null) {
                if (oldNode.kind !== newNode.kind)
                    wrappedNode.forget();
                else
                    wrappedNode.forgetDescendants();
            }
        }
```

As I noted in the comment, it is using an incorrect way of getting the start of the node and should be using .pos.

ofersadgat:
Fixes: https://github.com/dsherret/ts-morph/issues/1591

Note that using `.pos` matches what typescript is doing as well:

```
function visitNodes2(nodes, visitor, test, start, count) {
  if (nodes === void 0) {
    return nodes;
  }
  const length2 = nodes.length;
  if (start === void 0 || start < 0) {
    start = 0;
  }
  if (count === void 0 || count > length2 - start) {
    count = length2 - start;
  }
  let hasTrailingComma;
  let pos = -1;
  let end = -1;
  if (start > 0 || count < length2) {
    hasTrailingComma = nodes.hasTrailingComma && start + count === length2;
  } else {
    pos = nodes.pos;
    end = nodes.end;
    hasTrailingComma = nodes.hasTrailingComma;
  }
  const updated = visitArrayWorker(nodes, visitor, test, start, count); 
  if (updated !== nodes) {
    const updatedArray = factory.createNodeArray(updated, hasTrailingComma);
    setTextRangePosEnd(updatedArray, pos, end);
    return updatedArray;
  }
  return nodes;
}
```

The line:
```  const updated = visitArrayWorker(nodes, visitor, test, start, count); ```
is calling `visitArrayWorker` which calls `innerVisit` which first calls the transform function followed by the handleTransformation function. So, on the way out typescript is using `.pos` as well.

### DIFF
--- a/packages/ts-morph/src/compiler/ast/common/Node.ts
+++ b/packages/ts-morph/src/compiler/ast/common/Node.ts
@@ -1403,13 +1403,14 @@ export class Node<NodeType extends ts.Node = ts.Node> {
    * This will forget the current node and return a new node that can be asserted or type guarded to the correct type.
    * @param textOrWriterFunction - Text or writer function to replace with.
    * @returns The new node.
-   * @remarks This will replace the text from the `Node#getStart(true)` position (start position with js docs) to `Node#getEnd()`.
+   * @remarks This will replace the text from the `Node#getStart(true)` (or `Node#getPos()` if `replaceFullText` is true) position (start position with js docs) to `Node#getEnd()`.
    * Use `Node#getText(true)` to get all the text that will be replaced.
    */
   replaceWithText(textOrWriterFunction: string | WriterFunction): Node;
   /** @internal */
   replaceWithText(textOrWriterFunction: string | WriterFunction, writer: CodeBlockWriter): Node;
-  replaceWithText(textOrWriterFunction: string | WriterFunction, writer?: CodeBlockWriter): Node {
+  replaceWithText(textOrWriterFunction: string | WriterFunction, writer: CodeBlockWriter, replaceFullText: boolean): Node;
+  replaceWithText(textOrWriterFunction: string | WriterFunction, writer?: CodeBlockWriter, replaceFullText?: boolean): Node {
     const newText = getTextFromStringOrWriter(writer || this._getWriterWithQueuedIndentation(), textOrWriterFunction);
     if (Node.isSourceFile(this)) {
       this.replaceText([this.getPos(), this.getEnd()], newText);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by dsherret** [comment]
I'm not sure this is correct and it seems to cause some tests to fail.

**Comment by ofersadgat** [comment]
@dsherret Hi, I have done some more due diligence here. The code IS incorrect for a specific circumstance: JsxText. The reason for this is that getStart(true) includes comments but excludes leading whitespace. The problem, as you've mentioned elsewhere, is that leading whitespace is actually a part of a JsxText node. Therefore, excluding that when you're doing a transform is incorrect. This leads to the result of continually prepending whitespace to the transformed jsx node.

You run into a similar situation with replaceWithText(). This function similarly also doesnt include leading whitespace. This leads to the problem:
```
const text = node.getFullText();
node.replaceWithText(text);
```
this will also prepend whitespace which is counter intuitive. Also, afaict there isnt an api around this. So I have added an option into the replaceWithText api to allow replacing the full thing. If there is a prefered alternative change, lmk.

PR #1606, opened 2025-02-03
fix: update browser field to include fs/promises

### LINKED ISSUE #1604: browser support no longer works as of 24.0.0
**Describe the bug**

Version: >24.0.0

Since version 24.0.0 it seems browser support no longer works and throws this error when trying to load with an in-memory file system:

Image

It looks like this may have been introduced in #1549 [here](https://github.com/dsherret/ts-morph/pull/1549/files#diff-3a211c35604cae32fa6887c92239f844a1625adeeaacebdecea987f9de5e39f2).

**To Reproduce**

Using a framework like Next.js and trying to load ts-morph on the client will produce the error above:

```ts
'use client'
import { Project } from 'ts-morph'

export default function Page() {
  const fileSystem = new Project({
    useInMemoryFileSystem: true,
  })
  return null
}
```

**Expected behavior**

Similar to previous versions an in-memory file system should not error in browser environments.

souporserious:
Updates the browser field in `@ts-morph/common` `package.json` to include `fs/promises`.

fixes #1604

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -42,6 +42,7 @@
   },
   "browser": {
     "fs": false,
+    "fs/promises": false,
     "os": false,
     "fs.realpath": false,
     "mkdirp": false,


dsherret (approved):
Thanks and sorry this broke!

PR #1630, opened 2025-04-30
fix: Improve comma appending logic to handle nested template literals correctly

### LINKED ISSUE #1603: Adding method with useTrailingCommas modifies template strings unexpectedly
**Describe the bug**

Version: 25.0.0

It's a very specific scenario, but I'm reporting just in case there's something underlying that's more concerning. If you enable useTrailingCommas, adding a method to an object literal with a template string of the form `${p}/*` will add an extra comma to the template string `${p},/*`.

I might have time to try to dig deeper but figured I'd report it in case anything obvious sprung to mind for the root cause. (since it seems oddly specific for repro)

**To Reproduce**

```ts
import { Node, Project, SyntaxKind } from 'ts-morph';

const project = new Project({
  manipulationSettings: {
    useTrailingCommas: true,
  },
});
const sourceFile = project.createSourceFile('example.ts', 'func({});');

const callExpressions = sourceFile.getDescendantsOfKind(
  SyntaxKind.CallExpression,
);

const objectLiteral = callExpressions[0].getArguments()[0];

if (!Node.isObjectLiteralExpression(objectLiteral)) {
  throw new Error('Object literal expected');
}

objectLiteral.addMethod({
  name: 'logTemplate',
  statements: ['`${p}/*`;'],
  parameters: [{ name: 'p', type: 'string' }],
});

console.info(sourceFile.getFullText());
```

**Expected behavior**

```
func({
    logTemplate(p: string) {
        `${p}/*`;
    }
});
```

**Actual behavior**

```
func({
    logTemplate(p: string) {
        `${p},/*`;
    }
});
```

kingston:
Fixes #1603 

This PR fixes the comma‐appending logic so it correctly skips over nested template‐literal content (e.g. “${expr}/\*”) and appends commas in the right place. Previously, when inserting into comma-separated nodes, the scanner would misinterpret sequences like /\* inside a template literal as the start of a comment and place the comma inside the template literal incorrectly.

This PR mimics the logic found in classifier.ts (https://github.com/microsoft/TypeScript/blob/11e79327598db412a161616849041487673fadab/src/services/classifier.ts#L217) and keeps track of the template stack so that we skip any text values inside template literals using the `reScanTemplateToken` function of the scanner.

I also added a test to capture this case.

### DIFF
--- a/packages/ts-morph/src/manipulation/helpers/appendCommaToText.ts
+++ b/packages/ts-morph/src/manipulation/helpers/appendCommaToText.ts
@@ -21,14 +21,44 @@ export function getAppendCommaPos(text: string) {
   scanner.setText(text);
 
   try {
-    if (scanner.scan() === ts.SyntaxKind.EndOfFileToken)
+    let token = scanner.scan();
+
+    if (token === ts.SyntaxKind.EndOfFileToken)
       return -1;
 
-    while (scanner.scan() !== ts.SyntaxKind.EndOfFileToken) {
-      // just keep scanning...
+    // A stack to track nested template literals and their inner brace expressions
+    // otherwise, the scanner will interpret text in template literals as actual tokens,
+    // e.g. the /* in `${p}/*` will be interpreted as a comment, not inside the template literal
+    const templateStack: ts.SyntaxKind[] = [];
+
+    while (token !== ts.SyntaxKind.EndOfFileToken) {
+      switch (token) {
+        case ts.SyntaxKind.TemplateHead:
+          templateStack.push(token);
+          break;
+        case ts.SyntaxKind.OpenBraceToken:
+          if (templateStack.length > 0)
+            templateStack.push(token);
+          break;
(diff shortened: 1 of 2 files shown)


packages/ts-morph/src/manipulation/helpers/appendCommaToText.ts
```diff
@@ -21,14 +21,44 @@ export function getAppendCommaPos(text: string) {
...
+              templateStack.pop();
+          }
+          break;
+        }
+      }
+      token = scanner.scan();
     }
 
-    const pos = scanner.getStartPos();
+    const pos = scanner.getTokenFullStart();
```
kingston, line 61: Replaced this since getStartPos is deprecated and has been replaced by getTokenFullStart (that is identical in functionality) https://github.com/microsoft/TypeScript/blob/main/src/compiler/scanner.ts#L53

dsherret (approved):
Thanks!

Issue #1593, opened 2024-12-14
Adding `.d.ts` files into project no longer works in v24


B4nan:
**Describe the bug**

Version: 24.0.0

**To Reproduce**

We use ts-morph inside MikroORM to allow inference of property types from TS files. In production environment, we used `.d.ts` files instead of `.ts` to do this. This is the part that adds the files to the project:

https://github.com/mikro-orm/mikro-orm/blob/master/packages/reflection/src/TsMorphMetadataProvider.ts#L245

A complete reproduction for this problem is provided in https://github.com/mikro-orm/mikro-orm/issues/6297, I can try to narrow it down if that's not enough. Reverting ts-morph to v23 resolves the problem.

I've started debugging this from inside the docker container, and I can see that when we try to add the `.d.ts` files into the project via `this.sources = this.project.addSourceFilesAtPaths(paths)`, it's apparently a no-op, since calling `console.log(this.sources.map(s => s.getFilePath()))` gave me an empty array.

The inputs for `addSourceFilesAtPaths` are as follows:

```
[
  '/dist/shared/crud/crud.entity.d.ts',
  '/dist/apis/user/entities/user.entity.d.ts',
  '/dist/shared/crud/base.entity.d.ts',
  '/dist/apis/user/entities/test.entity.d.ts'
]
```

I verified all the files are present on those paths inside the container. The app is in the root of the container. Note that the error message you can see when you run that app says the `.ts` file was not found, but internally [we check both `.ts` and `.d.ts` files](https://github.com/mikro-orm/mikro-orm/blob/master/packages/reflection/src/TsMorphMetadataProvider.ts#L31-L33), and only report the former, so it's not about that.

I hope this can be enough for you since all the ts-morph handling is part of that single `TsMorphMetadataProvider` class (which is fairly simple), but I can surely try to reproduce this without the ORM involved too.

**Expected behavior**

Same behavior as v23, `.d.ts` files should be added to the project.

My hunch is it was caused by #1558. I tried to update TS to 5.6 and 5.7 in that project and that itself didn't help.

B4nan:
A bit more details about this, tried logging the inputs here:

https://github.com/dsherret/ts-morph/blob/061a3febe2383c4f0df32ac1339294dfe3f1f851/packages/ts-morph/src/fileSystem/DirectoryCoordinator.ts#L65

[...]

PR #1635, opened 2025-05-25
perf: switch to tinyglobby

benmccann:
It was confirmed in https://github.com/dsherret/ts-morph/issues/1593#issuecomment-2872190440 that the latest version of `tinyglobby` works

Since the last attempt, `tinyglobby` has now been adopted by `vite`, `pnpm`, `node-gyp`, `copy-webpack-plugin`, `storybook`, and many others. All of these additional users really help ensure that any bugs are very quickly identified and fixed

### DIFF
--- a/packages/common/src/runtimes/NodeRuntime.ts
+++ b/packages/common/src/runtimes/NodeRuntime.ts
@@ -1,9 +1,9 @@
-import fastGlob from "fast-glob";
 import * as minimatch from "minimatch";
 import * as fs from "node:fs";
 import * as fsp from "node:fs/promises";
 import * as os from "node:os";
 import * as path from "node:path";
+import { glob, globSync } from "tinyglobby";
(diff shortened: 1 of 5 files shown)


packages/common/src/runtimes/NodeRuntime.ts
```diff
@@ -163,14 +163,18 @@ class NodeRuntimeFileSystem implements RuntimeFileSystem {
   }
 
   glob(patterns: ReadonlyArray<string>) {
-    return fastGlob(patterns as string[], {
+    return glob({
+      patterns: patterns as string[],
```
SuperchupuDev: using `patterns` inside options is deprecated, although it's unlikely for a major tinyglobby version to happen anytime soon you should use it as first argument instead

benmccann: thanks. updated!

dsherret (approved):
Ok, let's try this again. Thanks!

PR #1637, opened 2025-05-29
fix: re-export StandardizedFilePath from @ts-morph/common

### LINKED ISSUE #1550: ts-morph package does not re-export types from @ts-morph/common
**Describe the bug**

Version: 26.0.0

If you are exporting types from your project that are using inferred types from `@ts-morph/common`, you'll get compiler errors when using package managers like `pnpm` which have stricter behavior around transitive dependencies.

**To Reproduce**

Create a typescript project that depends on `ts-morph` and export a function like this:

```ts
import { Project, SourceFile } from "ts-morph";

const file = project.createSourceFile("test.ts", ``);

export function getSourceFilePath() {
    return file.getFilePath() // Inferred return type is StandardizedFilePath from @ts-morph/common
}
```

If you're using a package manager like `pnpm`, compiling the project should produce an error like this:

```
src/Example.ts:6:17 - error TS2742: The inferred type of 'getSourceFilePath' cannot be named without a reference to '.pnpm/@ts-mor
ph+common@0.24.0/node_modules/@ts-morph/common'. This is likely not portable. A type annotation is necessary.

6  export function getSourceFilePath() {
                   ~~~~~~~~~~~~~~~~~
```

The only way to resolve this is to declare a direct dependency on `@ts-morph/common` which, as specified in that package's readme, you are NOT supposed to do and shouldn't need to.

**Expected behavior**

I should not need to depend directly on `@ts-morph/common` in order to export inferred types from `ts-morph`. The solution (as mentioned in [this thread](https://github.com/microsoft/TypeScript/issues/42873)) is to re-export types from `@ts-morph/common` in `ts-morph` so that that can be used directly by dependent projects.

kronodeus:
This resolves https://github.com/dsherret/ts-morph/issues/1550


dsherret (approved):
LGTM. Thanks!

PR #1654, opened 2025-09-17
fix: browser support no longer working since #1628 in 26.0.0 due to missing browser field

BossSloth:
#1628 replaced some imports to be prefixed with `node:` this broke browser support. This pr fixes that by added them to the browser package.json field. Both are needed due to mixed usages.


dsherret (approved):
Thanks!

BossSloth:
@dsherret looks like the publish failed due to no new version number https://github.com/dsherret/ts-morph/actions/runs/17808419850/job/50625875006
Could you publish the patch version to npm?

dsherret:
My bad! I missed this message. I just published a new version now.

